PR #82, opened 2025-11-23
Overhaul partition handling

rolandbeisel:


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
```diff
@@ -42,52 +44,66 @@ class OutboxProcessingScheduler(
...
     @Scheduled(fixedDelayString = "\${outbox.poll-interval:2000}")
     fun process() {
         try {
-            val myInstanceId = instanceRegistry.getCurrentInstanceId()
-            val assignedPartitions = partitionCoordinator.getAssignedPartitions(myInstanceId)
-
-            if (assignedPartitions.isEmpty()) {
-                log.debug("No partitions assigned to instance {} - waiting for rebalancing", myInstanceId)
-                return
+            if (rebalanceSignal.consume()) {
```
Alek96: This is a matter of personal preference, but I would move this check to `partitionCoordinator.rebalance`.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/NewInstancesRebalancer.kt
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param partitionContext current ownership & target distribution
+     * @param previousActiveInstanceIds instance IDs known in the last rebalance cycle
+     */
+    fun rebalance(
+        partitionContext: PartitionContext,
+        previousActiveInstanceIds: Set<String>,
+    ) {
+        val currentInstanceId = partitionContext.currentInstanceId
+        val newInstanceIds = partitionContext.activeInstanceIds - previousActiveInstanceIds
```
Alek96: I’m not entirely sure this check is necessary. It’s true that releasing partitions is pointless when no new instance was deployed, but the following check (`if (partitionsToReleaseCount == 0) return`) covers essentially the same scenario.
If `partitionsToReleaseCount <= 0`, then either:
- no new instance was deployed, 
or
- the number of deployed instances is less than or equal to the number of removed instances

Removing argument `previousActiveInstanceIds` would make the code cleaner.

Example:
```kotlin
fun rebalance(
    partitionContext: PartitionContext,
) {
    val currentInstanceId = partitionContext.currentInstanceId

    val partitionsToReleaseCount = partitionContext.countPartitionsToRelease()
    if (partitionsToReleaseCount == 0) return

    val partitionToRelease =
        partitionContext
            .getAllPartitionNumbersByInstanceId(currentInstanceId)
            .sorted()
            .takeLast(partitionsToReleaseCount)
            .toSet()
    if (partitionToRelease.isEmpty()) return

    try {
        partitionAssignmentRepository.releasePartitions(partitionToRelease, currentInstanceId)
    } catch (_: Exception) {
        log.debug("Could not release partitions {} from active instance {}", partitionToRelease, currentInstanceId)
        return
    }

    log.debug("Successfully released {} partitions of {}", partitionsToReleaseCount, currentInstanceId)
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/NewInstancesRebalancer.kt
```diff
@@ -0,0 +1,58 @@
...
+        val partitionsToReleaseCount =
+            max(0, partitionContext.countOwnedPartitionAssignments() - partitionContext.targetPartitionCount)
+        if (partitionsToReleaseCount == 0) return
+
+        val partitionToRelease =
+            partitionContext
+                .getAllPartitionNumbersByInstanceId(currentInstanceId)
+                .sorted()
+                .takeLast(partitionsToReleaseCount)
+                .toSet()
```
Alek96: You can add check `if (partitionToRelease.isEmpty()) return` to make it consistent with `StalePartitionsRebalancer`

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/NewInstancesRebalancer.kt
```diff
@@ -0,0 +1,58 @@
...
+    fun rebalance(
+        partitionContext: PartitionContext,
+        previousActiveInstanceIds: Set<String>,
+    ) {
+        val currentInstanceId = partitionContext.currentInstanceId
+        val newInstanceIds = partitionContext.activeInstanceIds - previousActiveInstanceIds
+        if (newInstanceIds.isEmpty()) return
+
+        val partitionsToReleaseCount =
+            max(0, partitionContext.countOwnedPartitionAssignments() - partitionContext.targetPartitionCount)
```
Alek96: Consider moving it to the `PartitionContext` class, e.g.,
```kotlin
fun countPartitionsToRelease(): Int = 
  max(0, countOwnedPartitionAssignments() - targetPartitionCount)

val partitionsToReleaseCount = partitionContext.countPartitionsToRelease()
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/PartitionAssignmentRepository.kt
```diff
@@ -0,0 +1,77 @@
...
+     * Concurrency:
+     * - Optimistic locking may raise exceptions under concurrent modifications (rollback). Caller can implement retry logic.
+     *
+     * @param partitionIds Set of partition numbers to claim.
+     * @param staleInstanceIds Set of stale instance IDs whose partitions may be taken over. May be null when only unassigned partitions should be claimed.
+     * @param newInstanceId The instance ID that will own these partitions after the operation.
+     */
+    fun claimStalePartitions(
+        partitionIds: Set<Int>,
+        staleInstanceIds: Set<String>?,
```
Alek96: With `staleInstanceIds` you are double-checking the same login as in `stalePartitionRebalancer`. 
In my opinion, you do not have to it. It is safe to do it once in `stalePartitionRebalancer`.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/PartitionCoordinator.kt
```diff
@@ -2,206 +2,114 @@ package io.namastack.outbox.partition
...
-        var newPartitions = 0
-
-        newAssignments.forEach { (partition, newInstance) ->
-            val oldInstance = oldAssignments[partition]
-            when {
-                oldInstance == null -> newPartitions++
-                oldInstance != newInstance -> transferred++
-            }
-        }
+        stalePartitionRebalancer.rebalance(partitionContext)
```
Alek96: Personal opinion: 
```kotlin
stalePartitionRebalancer.increasePartitionSize(partitionContext)
newInstanceRebalancer.decreasePartitionSize(partitionContext, lastKnownInstanceIds)
```
is more readable, than `rebalanace`

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/PartitionCoordinator.kt
```diff
@@ -2,206 +2,114 @@ package io.namastack.outbox.partition
...
-
-    /**
-     * Calculates optimal partition assignments for the given instances.
-     *
-     * @param sortedInstances List of instance IDs sorted consistently
-     * @return Map of partition to instance assignments
-     */
-    private fun calculatePartitionAssignments(sortedInstances: List<String>): Map<Int, String> {
-        val assignments = mutableMapOf<Int, String>()
+        val partitionContext =
```
Alek96: Extracting this to a private method could improve readability:
```kotlin
fun rebalance() {
    log.debug("Starting rebalance for instance {}", currentInstanceId)

    val partitionContext = createPartitionContext()

    if (partitionContext.activeInstanceIds.isEmpty()) {
        log.warn("No active instances found.")
        return
    }
    if (partitionContext.hasNoPartitionAssignments()) {
        bootstrapPartitions()
        return
    }

    stalePartitionRebalancer.increasePartitionSize(partitionContext)
    newInstanceRebalancer.decreasePartitionSize(partitionContext)
    cachedAssignedPartitions = null
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/StalePartitionsRebalancer.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Reclaim stale partitions up to the current shortage for this instance.
+     * @param partitionContext snapshot of current assignments and targets
+     */
+    fun rebalance(partitionContext: PartitionContext) {
+        val ownedPartitionCount = partitionContext.countOwnedPartitionAssignments()
+        val targetPartitionCount = partitionContext.targetPartitionCount
+
+        val partitionsToClaimCount = max(0, targetPartitionCount - ownedPartitionCount)
```
Alek96: Consider moving it to the `PartitionContext` class, e.g.,
```kotlin
fun countPartitionsToClaim(): Int = 
  max(0, targetPartitionCount - countOwnedPartitionAssignments())

val partitionsToClaimCount = partitionContext.countPartitionsToClaim()
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/StalePartitionsRebalancer.kt
```diff
@@ -0,0 +1,70 @@
...
+        if (partitionsToClaimCount == 0) return
+
+        val stalePartitionAssignments =
+            partitionContext.getStalePartitionAssignments().sortedBy { it.partitionNumber }
```
Alek96: The `.sortedBy { it.partitionNumber }` could be removed as `partitionsToClaim` is also being sort and `staleInstanceIds` is a set (the order is going to be changed)

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/StalePartitionsRebalancer.kt
```diff
@@ -0,0 +1,70 @@
...
+                .take(partitionsToClaimCount)
+                .toSet()
+
+        if (partitionsToClaim.isEmpty()) return
+
+        val staleInstanceIds = stalePartitionAssignments.mapNotNull { it.instanceId }.toSet().ifEmpty { null }
+        val currentInstanceId = partitionContext.currentInstanceId
+
+        try {
+            partitionAssignmentRepository.claimStalePartitions(
```
Alek96: If you decide to remove the `staleInstanceIds` argument from `partitionAssignmentRepository.claimStalePartitions` (`partitionAssignmentRepository.claimStalePartitions` duplicate logic from `StalePartitionsRebalancer.rebalance`), the code could become a little cleaner:
```kotlin
fun rebalance(partitionContext: PartitionContext) {
    val partitionsToClaimCount = partitionContext.countPartitionsToClaim()
    if (partitionsToClaimCount == 0) return

    val partitionsToClaim =
         partitionContext.getStalePartitionAssignments()
            .map { it.partitionNumber }
            .sorted()
            .take(partitionsToClaimCount)
            .toSet()
    if (partitionsToClaim.isEmpty()) return

    val currentInstanceId = partitionContext.currentInstanceId

    try {
        partitionAssignmentRepository.claimStalePartitions(
            partitionIds = partitionsToClaim,
            newInstanceId = currentInstanceId,
        )
    } catch (_: Exception) {
        log.debug("Could not claim partitions {} for current instance {}", partitionsToClaim, currentInstanceId)
        return
    }

    log.debug(
        "Successfully claimed {} partitions for current instance {}: {}",
        partitionsToClaimCount,
        currentInstanceId,
        partitionsToClaim,
    )
}
```

namastack-outbox-jpa/src/main/kotlin/io/namastack/outbox/JpaOutboxPartitionAssignmentRepository.kt
```diff
@@ -0,0 +1,177 @@
...
+     * @throws IllegalStateException if partition missing or owned by different instance
+     */
+    override fun claimStalePartitions(
+        partitionIds: Set<Int>,
+        staleInstanceIds: Set<String>?,
+        newInstanceId: String,
+    ) = transactionTemplate.executeNonNull {
+        val now = OffsetDateTime.now(clock)
+
+        val entities =
```
Alek96: A for loop could be better here:
```kotlin
override fun claimStalePartitions(
    partitionIds: Set<Int>,
    newInstanceId: String,
) = transactionTemplate.executeNonNull {
    val now = OffsetDateTime.now(clock)

    for (partitionNumber in partitionIds) {
        val entity = entityManager.find(OutboxPartitionAssignmentEntity::class.java, partitionNumber)
            ?: throw IllegalStateException("Partition cannot be claimed: missing",)
    
        entity.reassignTo(newInstanceId, now)
        entity.let { entityManager.merge(it) }
    }
}
```

Alek96 (approved):
I’ve added a few comments, but they’re mostly related to my personal coding style. Overall, it’s a +1 from me.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/NewInstancesRebalancer.kt
rolandbeisel: This is impossible and almost impossible to test, because of the code before.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/PartitionAssignmentRepository.kt
rolandbeisel: Passing `staleInstanceIds` to `claimStalePartitions` is essential for data consistency and race condition safety.
It ensures that only partitions still owned by stale or unassigned instances are claimed, even if another instance has already taken ownership between the stale scan and the claim attempt.
This double-check in the repository method is necessary because only the database transaction can guarantee the partition is still stale at the moment of claiming.

Alek96: I understand it now—you’re using `staleInstanceIds` to check whether the records have changed state. If another instance has taken over the partition, then the condition
`staleInstanceIds?.contains(entity.instanceId) != true && entity.instanceId != null`
evaluates to true because `entity.instanceId` is no longer in the `staleInstanceIds` set, which is similar to how optimistic locking works.
To make it even closer to optimistic locking semantics, you could rewrite it using a version field, for example:
```kotlin

override fun claimStalePartitions(
    partitions: Set<PartitionAssignment>,
    newInstanceId: String,
) = transactionTemplate.executeNonNull {
    val now = OffsetDateTime.now(clock)
    val query = """
        update OutboxPartitionAssignmentEntity p
        set p.instanceId = :cid, p.updatedAt = :ts, p.version = :v + 1
        where p.partition_number = :id and p.version = :v
    """

    for (partition in partitions) {
        val rowsUpdated =
            entityManager
                .createQuery(query)
                .setParameter("cid", newInstanceId)
                .setParameter("ts", now)
                .setParameter("partition_number", partition.partitionNumber)
                .setParameter("v", partition.version)
                .executeUpdate()

        if (rowsUpdated != 1) {
            throw IllegalStateException(
                "Partition cannot be claimed: either missing or owned by different instance",
            )
        }
    }
}
```
or if we want to use `OptimisticLockingFailureException`
```kotlin
override fun claimStalePartitions(
    partitions: Set<PartitionAssignment>,
    newInstanceId: String,
) = transactionTemplate.executeNonNull {
    val now = OffsetDateTime.now(clock)

    for (partition in partitions) {
         val entity = OutboxPartitionAssignmentEntityMapper.toEntity(partition)
         entity.reassignTo(newInstanceId, now)
         entityManager.persist(entity)
    }
}
```

But your approach with `staleInstanceIds` is perfectly fine.

Alek96: A question comes to my mind: why not be greedier?
Currently, we either assign all desired partitions or none (the transaction is fully committed or rolled back). Instead, why not allow assigning up to the desired number of partitions? We could attempt to reassign each partition to newInstanceId, and if one is already taken, simply skip it. This way, the instance would still increase its assigned partitions, even if only partially.

rolandbeisel: Because it would result in an unbalanced distribution of partitions, and achieving an even balance could require many runs. Also keep in mind that rebalancing is typically not something that happens regularly, at least in my experience.

rolandbeisel:
> I’ve added a few comments, but they’re mostly related to my personal coding style. Overall, it’s a +1 from me.

Hey @Alek96, thanks again for reviewing my code! I’ve done some overall refactoring, and I’d be really happy if you could find a moment to look over the changes. Hopefully most of the points from your last review are also addressed now.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/partition/PartitionAssignmentRepository.kt
rolandbeisel: I now use the (existing) version field for optimistic locking and pass in a `Set<PartitionAssignment>`. It's now very simple, which I like a lot :)

namastack-outbox-jpa/src/main/kotlin/io/namastack/outbox/JpaOutboxPartitionAssignmentRepository.kt
```diff
@@ -0,0 +1,120 @@
...
+     * - Once assignments are created by one instance, all other instances will see them
+     *   and skip their initialization attempts
+     * - No duplicate data is created despite concurrent startup
+     *
+     * @param partitionAssignments Set of partition assignments to insert if table is empty
+     * @return true if this instance initialized the assignments, false if another instance already did
+     */
+    override fun insertIfAllAbsent(partitionAssignments: Set<PartitionAssignment>): Boolean =
+        transactionTemplate.executeNonNull {
+            entityManager.find(OutboxPartitionLockEntity::class.java, 1, PESSIMISTIC_WRITE)
```
Alek96: I understand that we use this lock to ensure that only one instance can insert partitions into the database at a time. However, it feels like unnecessary overhead. We’re creating a separate table with a single record that’s only used once — during the initial deployment. On subsequent deployments, the second check (`val existing = entityManager.find(OutboxPartitionAssignmentEntity::class.java, 1)`) would work even without the lock.

The previous solution, relying on a primary-key constraint exception, seems like a better fit. Even if it triggers a commit rollback, it would happen only once during the entire lifetime of the application.

As a hacky alternative (and not my preferred one), we could insert the initial record into the `outbox_partitions` table during schema creation and then use a pessimistic lock on that row. To minimize the lock’s impact on the rest of the application, it might be worth reordering the lock and find operations (i.e., using a Double-Checked Locking pattern).
```kotlin
val existing = entityManager.find(OutboxPartitionAssignmentEntity::class.java, 1)
if (existing == null) {
    existing = entityManager.find(OutboxPartitionAssignmentEntity::class.java, 1, PESSIMISTIC_WRITE)
    if (existing == null) {
        ....
    }
}
```

What made you change the solution? I’d like to understand what I may have missed.

rolandbeisel: The switch to pessimistic locking was mainly to avoid error logs caused by unique constraint violations when multiple instances start up concurrently. While relying on the primary key constraint is technically safe, it results in unavoidable SQL errors in the logs (logged by Hibernate’s `SqlExceptionHelper` on ERROR level and you cannot prevent it), which some teams prefer to avoid for clarity and monitoring reasons. The lock table ensures only one instance initializes the partitions, and all others skip the insert without errors. If you’re comfortable with a one-time error log during startup, the original approach is perfectly valid and simpler. The locking solution is just a way to keep the logs clean and make the initialization process more explicit. But i have to admin that the lock table is overhead...I was also thinking about inserting a "bootstrap-instance" at schema initialization and lock this. After the partitions are initialized I could remove this instance in the same transaction. So exactly the same idea you had, but yeah it's kind of hacky...

rolandbeisel: After reconsidering, I decided to remove the pessimistic locking and accept the error log as a trade-off for simpler code and less database complexity. The primary key constraint reliably prevents duplicates, and the error only occurs once during initialization. Your suggestion matches my reasoning, and I now prefer the straightforward approach, even if it means living with that single error log.

Issue #89, opened 2025-11-28
Incremental Outbox Processing


Alek96:
Currently, during every outbox processing cycle, the library picks up to `outbox.batch-size` aggregate IDs to process in a single cycle. After all messages are sent, the cycle ends. But what if, instead of ending the cycle, we fetched another batch of messages? This way, we could process all undelivered messages before the next cycle.

The code could look like this:
```kotlin
@Scheduled(fixedDelayString = $$"${outbox.poll-interval:2000}", scheduler = "outboxDefaultScheduler")
fun process() {
    var previouslyAssignedPartitions: Set<Int>? = null

    while(true) {
        val assignedPartitions = partitionCoordinator.getAssignedPartitionNumbers()
        if (assignedPartitions.isEmpty()) break
        if (previouslyAssignedPartitions != null && assignedPartitions != previouslyAssignedPartitions) break
    
        val aggregateIds = recordRepository.findAggregateIdsInPartitions(assignedPartitions, ...)
        if (aggregateIds.isEmpty()) break
  
        val latch = CountDownLatch(aggregateIds.size)

        aggregateIds.forEach { aggregateId ->
            taskExecutor.execute {
                try {
                    processAggregate(aggregateId)
                } finally {
                    latch.countDown()
                }
            }
        }

        latch.await()
        previouslyAssignedPartitions =  assignedPartitions
    }
}
```

To improve this even further, we could fetch a new batch before the previous one is fully processed. This prevents wasting `outboxTaskExecutor` threads that would otherwise be idle while waiting for the next batch of messages.

Example scenario
1. Prepare `outboxTaskExecutor` with 4 threads
2. Trigger the outbox processing cycle
3. Fetch the 1st batch with 10 aggregate IDs
4. Process 5 out of 10 aggregates, leaving 5 remaining
5. Fetch the 2nd batch with 8 aggregate IDs (total now: 8 + 5)
6. Process 8 out of 13 aggregates, leaving 5 remaining
7. Fetch the 3rd batch with 0 aggregate IDs (total now: 5)
8. Wait for the remaining aggregates to be processed
9. End the outbox processing cycle

```kotlin
@Scheduled(fixedDelayString = $$"${outbox.poll-interval:2000}", scheduler = "outboxDefaultScheduler")
fun process() {
    var previouslyAssignedPartitions: Set<Int>? = null
    val limiter = OtboxProcessingLimiter(5) // user defined value threshold - e.g., 5

    while(true) {
[...]

PR #97, opened 2025-12-02
GH-89 add incremental outbox processing

Alek96:
Add incremental outbox processing with a rate limiter that enforces only `batch-size` recordKeys (e.g., 10) to be processed concurrently. Once we schedule `batch-size` tasks, we immediately fetch another batch.

---

**Notes to point out:**
- Now we fetch twice as many 'batch-size' to ensure we get enough new records even when some are still being processed
```kotlin
val recordKeys =
    recordRepository.findRecordKeysInPartitions(
        partitions = assignedPartitions,
        status = NEW,
        batchSize = properties.batchSize * 2, // <-----
        ignoreRecordKeysWithPreviousFailure = properties.processing.stopOnFirstFailure,
    )
```

- I'm not sure we need to check for rebalancing here. Currently, when rebalancing happens, we don't notify any of the processing recordKey threads to stop. Given that, why should we stop `OutboxProcessingScheduler`? The next batch will already use the updated partition list, so processing a new batch with the new partitions shouldn't break any logic.
```kotlin
if (assignedPartitions != initialAssignedPartitions) break
```

- I reduced the `outboxDefaultScheduler` `poolSize` to 1. Since the scheduler uses `fixedDelayString` and each job waits for all recordKeys to finish processing, only one job can run at a time. Using five threads offers no benefit, so a single thread is sufficient.

---

**Example with batch-size = 10:**
1. Trigger the outbox processing cycle
2. Fetch the 1st batch with 20 recordKeys
3. Schedule 10 tasks to process 10 recordKeys
4. As each task finishes, schedule a new task
5. Wait for 10 tasks to finish
6. Fetch the 2nd batch with 18 recordKeys (from which 10 are being processed, and 8 are new)
7. As each task finishes, schedule a new task
8. Wait for 10 tasks to finish
6. Fetch the 3rd batch with 10 recordKeys (from which 10 are being processed, and 0 are new)
7. Wait for the remaining recordKeys to be processed
8. End the outbox processing cycle

### DIFF
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
@@ -112,7 +112,6 @@ class OutboxProcessingScheduler(
 
             while (true) {
                 val assignedPartitions = partitionCoordinator.getAssignedPartitionNumbers()
-                if (assignedPartitions.isEmpty()) return
                 if (assignedPartitions != initialAssignedPartitions) break
 
                 val unprocessedRecordKeys = limiter.getUnprocessedIds()
--- a/namastack-outbox-core/src/test/kotlin/io/namastack/outbox/OutboxProcessingSchedulerTest.kt
+++ b/namastack-outbox-core/src/test/kotlin/io/namastack/outbox/OutboxProcessingSchedulerTest.kt
@@ -115,6 +115,19 @@ class OutboxProcessingSchedulerTest {
 
             verify(exactly = 1) { recordRepository.findRecordKeysInPartitions(setOf(1, 2), NEW, 100, true) }
         }
+
+        @Test
+        fun `stop processing records if assigned partitions changes`() {
+            val assignedPartitions = setOf(1, 3, 5)
+
+            every { partitionCoordinator.getAssignedPartitionNumbers() } returns
+                assignedPartitions andThen
+                assignedPartitions + 7
+
+            scheduler.process()
+
+            verify(exactly = 0) { recordRepository.findRecordKeysInPartitions(any(), any(), any(), true) }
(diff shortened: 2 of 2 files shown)


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingLimiter.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Acquires a permit for the given ID, blocking if the limit is reached.
+     * Adds the ID to the set of currently processing IDs.
+     *
+     * @param id The unique identifier for the task to acquire.
+     */
+    fun acquire(id: String) {
+        processingIds.add(id)
+        semaphore.acquire()
```
rolandbeisel, line 43: I think this should be the other way round? First aquire, then add.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
```diff
@@ -41,53 +40,64 @@ class OutboxProcessingScheduler(
...
                         } finally {
-                            latch.countDown()
+                            limiter.release(aggregateId)
                         }
                     }
                 }
 
-                latch.await()
-                log.debug("Finished processing {} aggregates", aggregateIds.size)
+                if (nextAggregateIds.size <= properties.batchSize) break
```
rolandbeisel: Why "<="? If nextAggregateIds.size == properties.batchSize then there could be more records in DB

Alek96: I was wondering why this wasn’t caught by the tests, and it turns out the bug only appears on the fourth consecutive batch.

Issue #98, opened 2025-12-02
Consistent naming for record key and record type


rolandbeisel:
aggregateId and eventType should be renamed to record key and record type, because it does not have to be an event and the key does not have to be the aggregateId

Issue #96, opened 2025-12-02
Rename retryCount to failCount


Alek96:
`OutboxRecord` has a property `retryCount` that is intended to represent the number of retry attempts made to process the record. However, the value is incremented on every processing failure, meaning that after the very first attempt, `retryCount` becomes `1`, even though no actual retry has occurred yet.

```kotlin
private fun handleFailure(
    record: OutboxRecord,
    ex: Exception,
) {
    val retriesExhausted = record.retriesExhausted(properties.retry.maxRetries)
    if (retriesExhausted || !retryPolicy.shouldRetry(ex)) {
        record.markFailed()
    } else {
        val delay = retryPolicy.nextDelay(record.retryCount)
        record.scheduleNextRetry(delay, clock)
        record.incrementRetryCount() // <---------------
        )
    }

    recordRepository.save(record)
}
``` 

We should consider renaming `retryCount` to `failCount` to more accurately reflect that the value represents the number of processing failures, not the number of retries.

Then, the calculation of exponential backoff retry would be as follows:
```kotlin
override fun nextDelay(failCount: Int): Duration {
    if (failCount == 0) {
        return 0
    }

    val delayMillis = (initialDelay.toMillis() * backoffMultiplier.pow(failCount - 1)).toLong()

    return Duration.ofMillis(minOf(delayMillis, maxDelay.toMillis()))
}
```

PR #99, opened 2025-12-02
GH-98 refactor to consistent naming of columns and properties

rolandbeisel:
GH-98 refactor to consistent naming of columns and properties
GH-96 rename retryCount to failureCount

### DIFF
--- a/namastack-outbox-core/src/test/kotlin/io/namastack/outbox/retry/JitteredRetryPolicyTest.kt
+++ b/namastack-outbox-core/src/test/kotlin/io/namastack/outbox/retry/JitteredRetryPolicyTest.kt
@@ -48,7 +48,7 @@ class JitteredRetryPolicyTest {
         val result = jitteredPolicy.nextDelay(1)
 
         verify { basePolicy.nextDelay(1) }
-        // Result should be between baseDelay and baseDelay + jitter
+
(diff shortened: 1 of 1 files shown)


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/retry/ExponentialBackoffRetryPolicy.kt
```diff
@@ -33,14 +33,14 @@ class ExponentialBackoffRetryPolicy(
...
      * The result is capped at the maximum delay.
      *
-     * @param retryCount The current retry count
+     * @param failureCount The number of failures that have occurred
      * @return Calculated delay duration, capped at maxDelay
      */
-    override fun nextDelay(retryCount: Int): Duration {
-        val delayMillis = (initialDelay.toMillis() * backoffMultiplier.pow(retryCount.toDouble())).toLong()
+    override fun nextDelay(failureCount: Int): Duration {
+        val delayMillis = (initialDelay.toMillis() * backoffMultiplier.pow(failureCount.toDouble())).toLong()
```
Alek96: `failureCount` is now incremented before the delay is calculated, so we need to subtract 1 when computing the delay:

```kotlin
val delayMillis = (initialDelay.toMillis() * backoffMultiplier.pow(failureCount - 1)).toLong()
```

PR #100, opened 2025-12-06
GH-90 Support multiple processors

rolandbeisel:
…s and annotations

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,821 +9,572 @@
 
 # Namastack Outbox for Spring Boot
 
-A robust Spring Boot library that implements the **Outbox Pattern** for reliable message publishing
-in distributed systems — built and maintained by [Namastack](https://outbox.namastack.io).
-This library ensures that domain events are published reliably, even in the face of system failures,
-by using transactional guarantees and hash-based partitioning.
+A robust Spring Boot library for **Java and Kotlin** projects that implements the **Transactional Outbox Pattern** for reliable record publishing
+in distributed systems. Ensures records are never lost through atomic persistence and automatic retry logic
+with handler-based processing and partition-aware horizontal scaling.
 
-## Disclaimer
+## Key Features
 
-**This project is currently undergoing significant changes.**
-The next release (0.4.0) will introduce breaking changes.
-Please note that existing integrations will need to be updated once the release is published.
+- ✅ **Transactional Atomicity**: Records saved in same transaction as domain data
+- ✅ **Automatic Retry**: Exponential backoff, fixed delay, or jittered policies
+- ✅ **Ordered Processing**: Records with same key processed sequentially
+- ✅ **Handler-Based**: Annotation-based or interface-based handler registration
+- ✅ **Horizontal Scaling**: Automatic partition assignment across instances
+- ✅ **Zero Message Loss**: Database-backed with at-least-once delivery
+- ✅ **Type-Safe Handlers**: Generic or typed handler support
+- ✅ **Built-in Metrics**: Micrometer integration for monitoring
+- ✅ **Flexible Payloads**: Store any type - events, commands, notifications, etc.
 
-## Features
+---
 
-- 🔄 **Transactional Outbox Pattern**: Ensures events are never lost
-- 🎯 **Hash-based Partitioning**: Automatic partition assignment for horizontal scaling
-- 🔁 **Automatic Retry**: Multiple retry policies with configurable strategies
-- 📊 **Event Ordering**: Guarantees event processing order per aggregate
-- ⚡ **High Performance**: Optimized for high-throughput scenarios
-- 🛡️ **Race Condition Safe**: Partition-based coordination prevents conflicts
-- 📈 **Horizontally Scalable**: Dynamic instance coordination and rebalancing
-- 🎯 **Zero Message Loss**: Database-backed reliability
-- 🎲 **Jitter Support**: Randomized delays to prevent thundering herd
-- 📊 **Built-in Metrics**: Comprehensive monitoring with Micrometer integration
+## 📖 Documentation
 
-## Quick Start
+For detailed information about features, configuration, and advanced topics, visit the **[complete documentation](https://outbox.namastack.io)**.
 
-### 1. Add Dependencies
+Quick links:
+- [API Reference (Javadoc)](https://javadoc.io/doc/io.namastack/namastack-outbox-core)
+- [GitHub Issues](https://github.com/namastack/namastack-outbox/issues)
+- [GitHub Discussions](https://github.com/namastack/namastack-outbox/discussions)
 
-Add the library to your `build.gradle.kts`:
+---
 
-```kotlin
+## 🚀 Quick Start (5 Minutes)
+
+### 1. Add Dependency
+
+```gradle
 dependencies {
-    implementation("io.namastack:namastack-outbox-starter-jpa:0.3.0")
+    implementation("io.namastack:namastack-outbox-starter-jpa:0.4.0")
 }
-```
-
-Or if you're using Maven, add to your `pom.xml`:
-
-```xml
-
-<dependency>
-  <groupId>io.namastack</groupId>
-  <artifactId>namastack-outbox-starter-jpa</artifactId>
-  <version>0.3.0</version>
-</dependency>
 ```
 
 ### 2. Enable Outbox
 
-Add the `@EnableOutbox` and `@EnableScheduling` annotations to your Spring Boot application:
-
 ```kotlin
 @SpringBootApplication
 @EnableOutbox
-@EnableScheduling  // Required for automatic event processing
-class YourApplication
+@EnableScheduling
+class Application
 
 fun main(args: Array<String>) {
-    runApplication<YourApplication>(*args)
+    runApplication<Application>(*args)
 }
 ```
 
-### 3. Configure Clock Bean
-
-The library requires a Clock bean for time-based operations. Configure your own Clock, if you
-don't want to use the default Clock bean from namastack-outbox:
-
-```kotlin
-@Configuration
-class OutboxConfiguration {
-
-    @Bean
-    fun clock(): Clock = Clock.systemUTC()
-}
-```
-
-### 4. Configure Database
-
-The library requires two database tables. You can enable automatic schema creation:
-
-```yaml
-outbox:
-  schema-initialization:
-    enabled: true
-```
-
-Or create the tables manually. You can look up the latest database schemas for all supported
-databases [here](https://github.com/namastack/namastack-outbox/tree/main/namastack-outbox-jpa/src/main/resources/schema).
-
-### 5. Create Event Processor
-
-Implement `OutboxRecordProcessor` to handle your events:
+### 3. Create a Handler
 
 ```kotlin
 @Component
-class MyEventProcessor(
-    private val objectMapper: ObjectMapper,
-    private val messagePublisher: MessagePublisher,
-) : OutboxRecordProcessor {
-    
-    private val objectMapper = ObjectMapper()
-
-    override fun process(record: OutboxRecord) {
-        when (record.eventType) {
-            "OrderCreatedEvent" -> handleOrderCreated(record)
-            "OrderUpdatedEvent" -> handleOrderUpdated(record)
-            "OrderCanceledEvent" -> handleOrderCanceled(record)
-            else -> logger.warn("Unknown event type: ${record.eventType}")
-        }
-    }
-
-    private fun handleOrderCreated(record: OutboxRecord) {
-        val event = objectMapper.readValue(record.payload, OrderCreatedEvent::class.java)
-        messagePublisher.publish(event)
+class OrderRecordHandler : OutboxTypedHandler<OrderCreatedRecord> {
+    override fun handle(payload: OrderCreatedRecord) {
+        // Process the record - this will be called automatically when ready
+        eventPublisher.publish(payload)
     }
 }
 ```
 
-### 6. Save Outbox Records to Outbox
-
-Namastack Outbox supports two main ways to save events to the outbox table, both ensuring that your events are reliably persisted and later processed in the correct order:
-
-#### 1. Save via @OutboxEvent Annotation (Recommended, since 0.3.0)
-
-Annotate your event class with `@OutboxEvent` to enable automatic outbox persistence. The annotation supports SpEL expressions for dynamic aggregateId extraction and allows you to optionally specify the event type (otherwise, the fully qualified class name is used by default):
-
-```kotlin
-@OutboxEvent(
-    aggregateId = "#root.id.toString()", // required - supports SpEL expressions
-    eventType = "CustomerRegisteredEvent" // optional - default is qualifiedName of class
-)
-data class CustomerRegisteredEvent(
-    val id: UUID,
-    val firstname: String,
-    val lastname: String,
-    val email: String,
-)
-```
-
-To publish your event, simply use the standard Spring `ApplicationEventPublisher`:
-
-```kotlin
-applicationEventPublisher.publish(event)
-```
-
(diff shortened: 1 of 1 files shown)


namastack-outbox-jackson/src/main/kotlin/io/namastack/outbox/OutboxJacksonAutoConfiguration.kt
```diff
@@ -28,11 +28,12 @@ class OutboxJacksonAutoConfiguration {
     fun jsonMapper(): JsonMapper = JsonMapper.builder().build()
```
Alek96: `JsonMapper` is automatically configured by Spring Boot if Jackson is on the classpath. [Documentation](https://docs.spring.io/spring-boot/reference/features/json.html#features.json.jackson)
For now, it is fine, but in the future, we could think of using `JsonMapper` created by Spring

namastack-outbox-jackson/src/test/kotlin/io/namastack/outbox/JacksonEventOutboxSerializerTest.kt
```diff
@@ -10,7 +10,7 @@ import tools.jackson.databind.json.JsonMapper
 
 class JacksonEventOutboxSerializerTest {
     private val mapper = mockk<JsonMapper>()
```
Alek96: It would be better to use a real JsonMapper implementation instead of a mock.
```kotlin
private val mapper = JsonMapper()
```
Note: It is not relevant for this PR

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxRecord.kt
```diff
@@ -126,56 +127,55 @@ class OutboxRecord internal constructor(
...
 
         /**
          * Builds the OutboxRecord with the configured values.
          *
          * @param clock Clock to use for timestamps (defaults to system UTC)
          * @return A new OutboxRecord instance
          */
-        fun build(clock: Clock): OutboxRecord {
+        fun build(clock: Clock): OutboxRecord<T> {
+            val id = UUID.randomUUID().toString()
```
Alek96, line 160: `UUID.randomUUID().toString()` is used in 3 places to generate a random key:
- `OutboxEventMulticaster.saveOutboxRecord`
- `OutboxRecord.Builder.build`
- `OutboxService.schedule`

We could extract it, for example, to the `OutboxRecord` static method, to avoid duplicating the code on how a random key is generated.
```kotlin
@JvmStatic
fun randomKey(): String = UUID.randomUUID().toString()
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerRegistry.kt
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Retrieves all typed handlers that match a specific payload type.
+     *
+     * Returns an empty list if no typed handlers are registered for the type.
+     * Multiple handlers can be registered for the same type (all will be invoked).
+     *
+     * @param type The payload type to search for
+     * @return List of TypedHandlerMethods for this type (empty if none)
+     */
+    fun getHandlersForPayloadType(type: KClass<*>): List<TypedHandlerMethod> = typedHandlers[type] ?: emptyList()
```
Alek96: You are returning a mutable list here. To fix it, we could add `.toList()`
```kotlin
fun getHandlersForPayloadType(type: KClass<*>): List<TypedHandlerMethod> = typedHandlers[type]?.toList() ?: emptyList()
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerRegistry.kt
```diff
@@ -0,0 +1,119 @@
...
+     * Registers a handler in the global ID map.
+     *
+     * Ensures no duplicate handler IDs are registered - each handler method
+     * must have a unique ID derived from its class, method name, and parameters.
+     *
+     * @param handlerMethod The handler method to register globally
+     * @throws IllegalStateException if a handler with the same ID is already registered
+     */
+    private fun registerInAllHandlers(handlerMethod: OutboxHandlerMethod) {
+        if (handlersById.putIfAbsent(handlerMethod.id, handlerMethod) != null) {
```
Alek96: SonarQube code smell [[kotlin:S6532](https://next.sonarqube.com/sonarqube/coding_rules?open=kotlin:S6532&rule_key=kotlin:S6532)]: 
Replace this if expression with:
```kotlin
check(handlersById.putIfAbsent(handlerMethod.id, handlerMethod) == null) { 
    "Duplicate handler ID detected: ${handlerMethod.id}" 
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
```diff
@@ -0,0 +1,107 @@
...
+     *
+     * @param bean The newly instantiated bean
+     * @param beanName The name of the bean in the Spring context
+     * @return The original bean unchanged
+     */
+    override fun postProcessAfterInitialization(
+        bean: Any,
+        beanName: String,
+    ): Any {
+        // Scan for handlers using all configured scanners
```
Alek96: We could adapt something like a visitor design pattern here:
```kotlin
    val bean: Any,
    val method: Method,
) {
    abstract fun register(registry: OutboxHandlerRegistry)
}

class GenericHandlerMethod(
    bean: Any,
    method: Method,
) : OutboxHandlerMethod(bean, method) {

    override fun register(registry: OutboxHandlerRegistry) {
        registry.registerGenericHandler(handlerMethod = this)
    }
}

class TypedHandlerMethod(
    bean: Any,
    method: Method,
    val paramType: KClass<*>,
) : OutboxHandlerMethod(bean, method) {

    override fun register(registry: OutboxHandlerRegistry) {
        registry.registerTypedHandler(
            handlerMethod = this,
            paramType = paramType,
        )
    }
}
```
Then the `postProcessAfterInitialization` method would look like this:
```kotlin
override fun postProcessAfterInitialization(
    bean: Any,
    beanName: String,
): Any {
    // Scan for handlers using all configured scanners
    scanners
        .flatMap { it.scan(bean) }
        .forEach { handlerMethod -> handlerMethod.register(registry) }

    return bean
}
```

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/annotation/OutboxEvent.kt
```diff
@@ -1,20 +1,18 @@
...
  *
  * @author Roland Beisel
  * @since 0.3.0
  */
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.RUNTIME)
 annotation class OutboxEvent(
-    val key: String,
-    val eventType: String = "",
+    val key: String = "",
```
Alek96, line 17: Just an idea, if we are going to create `OutboxRecord` static method for key generation, then we could use this  method as a default value here:
```kotlin
val key: String = "T(io.namastack.outbox.OutboxRecord).randomKey()",
```

Then the `key` attribute could be mandatory.

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/handler/OutboxHandler.kt
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * If both a typed handler ([OutboxTypedHandler<T>]) and this generic handler
+ * are registered for the same payload type, both will be called:
+ * 1. Typed handler is invoked first (if payload type matches)
+ * 2. Generic handler is invoked second (always)
+ *
+ * @author Roland Beisel
+ * @since 0.4.0
+ */
+interface OutboxHandler {
```
Alek96, line 33: Consider renaming `OutboxHandler` to `OutboxGenericHandler` for consistency. This would align it with the existing naming pattern:
- `OutboxGenericHandler`, `GenericHandlerMethod`, `GenericHandlerMethodFactory`
- `OutboxTypedHandler`, `TypedHandlerMethod`, `TypedHandlerMethodFactory`

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/TypedHandlerMethod.kt
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @param bean The bean instance containing the handler method
+ * @param method The actual method to invoke (reflection)
+ * @param paramType The specific payload type this handler processes
+ */
+class TypedHandlerMethod(
+    bean: Any,
+    method: Method,
+    val paramType: KClass<*>,
+) : OutboxHandlerMethod(bean, method) {
```
Alek96, line 30: Future extension idea, to make `TypedHandlerMethod` not only work for
```kotlin
@OutboxHandler
fun handle(payload: String) {}
```
but also for
```kotlin
@OutboxHandler
fun handle(payload: String, metadata: OutboxRecordMetadata) {}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerInvoker.kt
```diff
@@ -0,0 +1,61 @@
...
+        metadata: OutboxRecordMetadata,
+    ) {
+        if (payload == null) return
+
+        val handler =
+            handlerRegistry.getHandlerById(metadata.handlerId)
+                ?: throw IllegalStateException("No handler with id ${metadata.handlerId}")
+
+        // Invoke handler based on type (typed vs generic)
+        when (handler) {
```
Alek96, line 56: If we make the `invoke` method an abstract method on `OutboxHandlerMethod`, then we would simplify the usage:
```kotlin
handler.invoke(payload, metadata)
```

namastack-outbox-integration-tests/src/test/kotlin/io/namastack/outbox/UnorderedProcessingIntegrationTest.kt
```diff
@@ -82,28 +85,31 @@ class UnorderedProcessingIntegrationTest {
...
-                .Builder()
-                .recordKey(recordKey)
+                .Builder<String>()
+                .key(recordKey)
                 .payload(payload)
-                .recordType("recordType")
-                .build(clock),
+                .handlerId(
+                    @Suppress("ktlint:standard:max-line-length")
+                    $$"io.namastack.outbox.OrderedProcessingIntegrationTest$TestProcessor#handle(java.lang.Object,io.namastack.outbox.handler.OutboxRecordMetadata)",
```
Alek96: Auto-detected error:
The handler ID references `OrderedProcessingIntegrationTest` but this is in `UnorderedProcessingIntegrationTest`. This appears to be a copy-paste error. The handler ID should reference the correct class name to ensure proper handler resolution at runtime.

Auto-generated fix
`TestProcessor::class.java.name + "#handle(java.lang.Object,io.namastack.outbox.handler.OutboxRecordMetadata)",`

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/TypedHandlerMethod.kt
rolandbeisel: Why should you need this? I have no use case in mind. And payload can be a String, yes... but normally it would be a record / data class of any kind. Metadata was introduced to have the key information when using generic OutboxHandlers. In the case of a specific outboxHandler like you showed, you will normally know which property you used as the key. Or do I oversee something?

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/handler/OutboxHandler.kt
rolandbeisel: Yes, I thought about that, too. But this will be the most used annotation imo and that's why i chose a very simple annotation name, even if it's incosistent.

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/annotation/OutboxEvent.kt
rolandbeisel: under the hood the key is mandatory (see OutboxRecord.Builder().build())... in your solution, somebody could set key = "".

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerInvoker.kt
rolandbeisel: Then both methods have to accept also the metadata param. See other discussion.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxRecord.kt
rolandbeisel: hmm this does not belong to an OutboxRecord... I was thinking about Value Objects for Ids, where this yould take place (e.g. OutboxRecordKey.create()). The other alternative would be to implement a Utitlity class for that, but UUID itself is exactly that.

Alek96: I agree that we shouldn’t introduce a new utility class just to duplicate UUID generation. The intent was simply to centralize the creation of new random keys. The current approach is fine as it is.

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/annotation/OutboxEvent.kt
Alek96: I initially considered using SpEL to indicate to the user that the default key value is a UUID. However, thinking about it now, adding this information to the documentation would make it clearer and more readable.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/TypedHandlerMethod.kt
Alek96: In this example, `String` was just an example. My idea was to allow users to include `OutboxRecordMetadata` as a second parameter in methods annotated with `OutboxHandler`. But, as you mentioned, I don’t have a concrete use case for it right now, so we can drop the idea for the time being. If someone reaches out requesting this capability, we can consider implementing it then.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerInvoker.kt
Alek96: You're right. If we added a `metadata` parameter, the implementation in `TypedHandlerMethod` would end up with an unused argument, which would still be a code smell. This really comes down to personal preference—when I see a `when` statement with different behavior for each implementation, I usually try to find an alternative design. But in this case, adding the extra parameter doesn’t improve things.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
rolandbeisel: Love that!

mmiesiak:
Thank you for your prompt action.

It might be helpful to highlight the difference in both error handling scenarios and clarify that using a custom retry policy and returning false in the ‘shouldRetry’ method will mark the task as FAILED, while not throwing an exception in the handler will complete the task successfully.

It wasn’t immediately apparent from examining both examples which option is preferred and when we should use each one.

I’ve been looking forward very much for 0.4.0 release.

Keep the good work.

Issue #108, opened 2025-12-09
Support TaskDecorator bean auto-configuration


Alek96:
Just as Spring Boot automatically applies `TaskDecorator` beans to its auto-created `ThreadPoolTaskExecutor` instances, we should do the same for our `ThreadPoolTaskExecutor` and `ThreadPoolTaskScheduler`. This would make it much easier for users who need to propagate trace context in asynchronous tasks executed by these executors.

Issue #106, opened 2025-12-08
Issue with type based handlers after upgrade to 0.4.0


mmiesiak:
This one was pulling my hair from my head yesterday night, when I was trying upgrade from previous version.

I started migration to `0.4.0` using annotation based handlers.

I had configured my handler like this

```
@Observed
@OutboxHandler
public void handle(ProfileInfoValueUpdatedEvent payload) {
    log.info("Processing ProfileInfoValueUpdatedEvent event: {}", payload);
}
```

After starting the application, no events were stored to outbox repository and there was no trace of any reason in the logs. 

After switching to typed handlers using an interface, the application failed at startup, and logs were more helpful:

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'profileInfoValueUpdatedEventProcessor' defined in file [/Users/dev/projects/demo/target/classes/com/demo/payroll/ProfileInfoValueUpdatedEventProcessor.class]: Collection contains no element matching the predicate.
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:610) ~[spring-beans-7.0.1.jar:7.0.1]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:525) ~[spring-beans-7.0.1.jar:7.0.1]
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:333) ~[spring-beans-7.0.1.jar:7.0.1]
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:371) ~[spring-beans-7.0.1.jar:7.0.1]
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:331) ~[spring-beans-7.0.1.jar:7.0.1]
	at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:196) ~[spring-beans-7.0.1.jar:7.0.1]
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.instantiateSingleton(DefaultListableBeanFactory.java:1218) ~[spring-beans-7.0.1.jar:7.0.1]
[...]

PR #109, opened 2025-12-09
Feature/auto configure task decorator

Alek96:
* Add support for proxy beans, inherited classes, and implemented interfaces for the `@OutboxHandler` annotation and the `OutboxTypedHandler` interface (fix GH-106).
* Add support for `TaskDecorator` (resolve GH-108).


rolandbeisel:
> @rolandbeisel I will split this PR into two

You don't need to. All good!

PR #118, opened 2025-12-16
GH-101 add handler specific retry policies

rolandbeisel:


namastack-outbox-api/src/main/kotlin/io/namastack/outbox/annotation/OutboxRetryable.kt
```diff
@@ -0,0 +1,26 @@
...
+ *     }
+ * }
+ * ```
+ *
+ * @property value Spring bean name of the OutboxRetryPolicy to use
+ */
+@Target(AnnotationTarget.FUNCTION)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class OutboxRetryable(
+    val value: String,
```
Alek96: I think it would be a good idea to give the possibility to specify the `OutboxRetryPolicy` bean not only by name, but also by class.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/retry/OutboxRetryPolicyRegistry.kt
```diff
@@ -0,0 +1,99 @@
...
+     * If the bean is not found, throws an exception with a helpful message
+     * listing all available retry policy beans.
+     *
+     * @param beanName The Spring bean name of the retry policy
+     * @return The loaded OutboxRetryPolicy bean
+     * @throws IllegalStateException if the bean is not found or has wrong type
+     */
+    fun loadRetryPolicyBean(beanName: String): OutboxRetryPolicy =
+        try {
+            beanFactory.getBean(beanName, OutboxRetryPolicy::class.java)
```
Alek96: Java-style call can be replaced with Kotlin extension:
```kotlin
beanFactory.getBean<OutboxRetryPolicy>(beanName)
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/retry/OutboxRetryPolicyRegistry.kt
```diff
@@ -0,0 +1,99 @@
...
+        } catch (ex: Exception) {
+            // List available policies for better error messages
+            val available =
+                if (beanFactory is ListableBeanFactory) {
+                    beanFactory.getBeansOfType(OutboxRetryPolicy::class.java).keys.sorted()
```
Alek96: Java-style call can be replaced with Kotlin extension:
```kotlin
beanFactory.getBeansOfType<OutboxRetryPolicy>().keys.sorted()
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
```diff
@@ -82,7 +88,10 @@ internal class OutboxHandlerBeanPostProcessor(
     ): Any {
         scanners
             .flatMap { it.scan(bean) }
-            .forEach { handlerMethod -> handlerMethod.register(registry) }
+            .forEach { handlerMethod ->
```
Alek96: Consider renaming the `handlerMethod.register` method to something like `handlerMethod.registerHandler` or `handlerMethod.selfRegister`.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/OutboxHandlerMethod.kt
```diff
@@ -58,4 +80,34 @@ sealed class OutboxHandlerMethod(
...
+        val annotation = AnnotatedElementUtils.findMergedAnnotation(method, OutboxRetryable::class.java)
+        if (annotation != null) {
+            return registry.loadRetryPolicyBean(annotation.value)
+        }
+
+        if (bean is OutboxRetryAware) {
+            return bean.getRetryPolicy()
+        }
+
+        return registry.getDefaultRetryPolicy()
```
Alek96: I’m not sure if this is a good idea, but I’ll share it anyway.
Perhaps it would be better not to register a `defaultRetryPolicy` and instead rely on the fallback behavior in `OutboxRetryPolicyRegistry`?
```kotlin
fun getByHandlerId(handlerId: String): OutboxRetryPolicy = policiesById[handlerId] ?: defaultRetryPolicy
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
```diff
@@ -82,7 +88,10 @@ internal class OutboxHandlerBeanPostProcessor(
...
+                handlerMethod.register(handlerRegistry)
```
Alek96, line 92: Consider renaming the `handlerMethod.register` method to something like `handlerMethod.registerHandler` or `handlerMethod.selfRegister`.

rolandbeisel: I will leave it for the moment, because i don't like the alternative namings and i don't come up with a good one.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/retry/OutboxRetryPolicyRegistry.kt
```diff
@@ -0,0 +1,130 @@
...
+     * throws an exception with a helpful message.
+     *
+     * @param policyClass The retry policy class to look up
+     * @return The loaded OutboxRetryPolicy bean
+     * @throws IllegalStateException if the bean is not found, multiple beans exist, or has wrong type
+     */
+    fun getRetryPolicy(policyClass: KClass<out OutboxRetryPolicy>): OutboxRetryPolicy =
+        try {
+            beanFactory.getBean(policyClass.java)
+        } catch (ex: Exception) {
```
Alek96, line 114: I wonder if extracting the code for simplicity here is a good call:
```kotlin
fun getRetryPolicy(beanName: String): OutboxRetryPolicy =
    try {
        beanFactory.getBean<OutboxRetryPolicy>(beanName)
    } catch (ex: Exception) {
        val available = getAvailableBeans()

        throw IllegalStateException(
            "Retry policy bean '$beanName' not found. Available: $available",
            ex,
        )
    }

fun getRetryPolicy(policyClass: KClass<out OutboxRetryPolicy>): OutboxRetryPolicy =
    try {
        beanFactory.getBean(policyClass.java)
    } catch (ex: Exception) {
        val available = getAvailableBeans()
        throw IllegalStateException(
            "Retry policy bean of type '${policyClass.simpleName}' not found. Available: $available",
            ex,
        )
    }

private fun getAvailableBeans(): String {
    val available =
        if (beanFactory is ListableBeanFactory) {
            beanFactory.getBeansOfType<OutboxRetryPolicy>().mapValues { it.value::class.simpleName }
        } else {
            emptyMap()
        }

    return available.entries.joinToString { "${it.key} (${it.value})" }
}
```

Issue #127, opened 2025-12-20
Add Fallback Handler Support for Failed Outbox Records


rolandbeisel:
Add support for fallback handlers that are automatically invoked when outbox record processing fails after all retry attempts are exhausted or when a non-retryable exception occurs. This enables graceful error handling through dead letter queues, compensation logic, alerting, and audit logging.

Two approaches for registering fallback handlers:

**Interface-Based**

```kotlin
@Component
class OrderHandler : OutboxTypedHandlerWithFallback<OrderEvent> {
    override fun handle(payload: OrderEvent) {
        emailService.send(payload)
    }

    override fun handleFailure(
        payload: OrderEvent,
        metadata: OutboxRecordMetadata,
        context: OutboxFailureContext
    ) {
        deadLetterQueue.publish(payload)
        alertingService.alert("Failed", context)
    }
}
```

**Annotation-Based**

```kotlin
@Component
class OrderHandlers {
    @OutboxHandler
    fun handleOrder(payload: OrderEvent) {
        emailService.send(payload)
    }

    @OutboxFallbackHandler
    fun handleOrderFailure(
        payload: OrderEvent,
        metadata: OutboxRecordMetadata,
        context: OutboxFailureContext
    ) {
        deadLetterQueue.publish(payload)
    }
}
```

**Key Features**

- Automatic type matching: Fallbacks matched to handlers by payload type
- Hierarchical matching: Exact type match first, then generic Any fallback as catch-all
- Shared fallbacks: Multiple handlers can share one fallback
- Failure context: Includes retriesExhausted, failureCount, lastFailureReason, etc.

PR #128, opened 2025-12-20
GH-127 support for fallback handlers

rolandbeisel:


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxRecordProcessor.kt
```diff
@@ -0,0 +1,182 @@
...
+     */
+    private fun handlePermanentFailure(
+        record: OutboxRecord<*>,
+        retriesExhausted: Boolean,
+        nonRetryable: Boolean,
+    ): Boolean {
+        val fallbackSuccess = invokeFallbackHandler(record, retriesExhausted, nonRetryable)
+
+        if (fallbackSuccess) {
+            record.markCompleted(clock)
```
Alek96: If the record is marked as completed and `properties.processing.deleteCompletedRecords` is set to true, should we delete the record?

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxRecordProcessor.kt
```diff
@@ -0,0 +1,182 @@
...
+            log.info(
+                "Record {} for key {} marked as COMPLETED after fallback handler succeeded ({} failures)",
+                record.id,
+                record.key,
+                record.failureCount,
+            )
+        } else {
+            record.markFailed()
```
Alek96: Should we save the fallback exception as record failureReason?
`record.updateFailureReason(ex.message)`

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxRecordProcessor.kt
```diff
@@ -0,0 +1,182 @@
...
+) {
+    private val log = LoggerFactory.getLogger(OutboxRecordProcessor::class.java)
+
+    /**
+     * Processes a single record through its handler.
+     *
+     * @param record The record to process
+     * @return true if record processed successfully (or fallback succeeded), false otherwise
+     */
+    fun processRecord(record: OutboxRecord<*>): Boolean =
```
Alek96: Personal taste - it is somehow hard to notice that the `handleFailure(record, ex)` returns the value.
Consider changing the method to a less functional style:
```kotlin
fun processRecord(record: OutboxRecord<*>): Boolean {
try {
        ....
        return true
    } catch (ex: Exception) {
        return handleFailure(record, ex)
    }
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxRecordProcessor.kt
```diff
@@ -0,0 +1,182 @@
...
+ * @param handlerInvoker Dispatcher for normal handlers
+ * @param fallbackHandlerInvoker Dispatcher for fallback handlers
+ * @param retryPolicyRegistry Registry for handler-specific retry policies
+ * @param properties Configuration properties
+ * @param clock Clock for time calculations
+ *
+ * @author Roland Beisel
+ * @since 0.5.0
+ */
+class OutboxRecordProcessor(
```
Alek96: I feel like there is something wrong with this class code. I tried to refactor it, and I came up with this:
(PS. `handleHandler` is not the best name) 

```kotlin
class OutboxRecordProcessor(
    private val recordRepository: OutboxRecordRepository,
    private val handlerInvoker: OutboxHandlerInvoker,
    private val fallbackHandlerInvoker: OutboxFallbackHandlerInvoker,
    private val retryPolicyRegistry: OutboxRetryPolicyRegistry,
    private val properties: OutboxProperties,
    private val clock: Clock,
) {
    private val log = LoggerFactory.getLogger(OutboxRecordProcessor::class.java)

    fun processRecord(record: OutboxRecord<*>): Boolean {
        val success = handleHandler(record)

        if (success && properties.processing.deleteCompletedRecords) {
            log.trace("Deleting record {} after successful processing", record.id)
            recordRepository.deleteById(record.id)
        } else {
            recordRepository.save(record)
        }
        return success
    }

    private fun handleHandler(record: OutboxRecord<*>): Boolean {
        try {
            log.trace("Processing record {} (key: {}, partition: {})", record.id, record.key, record.partition)

            val metadata = createRecordMetadata(record)
            handlerInvoker.dispatch(record.payload, metadata)

            log.trace("Marking record {} as COMPLETED", record.id)
            record.markCompleted(clock)
            return true
        } catch (ex: Exception) {
            return handleFailure(record, ex)
        }
    }

    private fun handleFailure(
        record: OutboxRecord<*>,
        ex: Exception,
    ): Boolean {
        log.debug("Failed processing record {} for key {}: {}", record.id, record.key, ex.message)

        record.incrementFailureCount()
        record.updateFailureReason(ex.message)

        val retryPolicy = retryPolicyRegistry.getByHandlerId(record.handlerId)
        val retriesExhausted = record.retriesExhausted(retryPolicy.maxRetries())
        val retryable = retryPolicy.shouldRetry(ex)

        return when {
            // can retry
            !retriesExhausted && retryable -> {
                scheduleRetry(record, retryPolicy)
                false
            }

            // cannot retry - has fallback
            fallbackHandlerInvoker.hasFallback(record.handlerId) -> {
                handleFallback(record, retriesExhausted, !retryable)
            }

            // cannot retry - no fallback
            else -> {
                markRecordAsFailed(record, retriesExhausted)
                false
            }
        }
    }

    private fun scheduleRetry(
        record: OutboxRecord<*>,
        retryPolicy: OutboxRetryPolicy,
    ) {
        val delay = retryPolicy.nextDelay(record.failureCount)
        record.scheduleNextRetry(delay, clock)

        log.debug("Scheduled retry #{} for record {} in {}", record.failureCount, record.id, delay)
    }

    private fun handleFallback(
        record: OutboxRecord<*>,
        retriesExhausted: Boolean,
        nonRetryable: Boolean,
    ): Boolean {
        try {
            log.trace("Processing record {} fallback (key: {}, partition: {})", record.id, record.key, record.partition)

            val metadata = createRecordMetadata(record)
            val context = createFailureContext(record, retriesExhausted, nonRetryable)
            fallbackHandlerInvoker.dispatch(record.payload, metadata, context)

            log.info("Marking record {} as COMPLETED after fallback handler succeeded", record.id)
            record.markCompleted(clock)
            return true
        } catch (ex: Exception) {
            log.error("Fallback handler failed for record {}: {}", record.id, ex.message, ex)
            markRecordAsFailed(record, retriesExhausted)
            return false
        }
    }

    private fun markRecordAsFailed(
        record: OutboxRecord<*>,
        retriesExhausted: Boolean,
    ) {
        record.markFailed()
        log.warn(
            "Record {} for key {} marked as FAILED after {} failures{}",
            record.id,
            record.key,
            record.failureCount,
            if (retriesExhausted) " (retries exhausted)" else " (non-retryable exception)",
        )
    }

    private fun createRecordMetadata(record: OutboxRecord<*>): OutboxRecordMetadata =
        OutboxRecordMetadata(
            key = record.key,
            handlerId = record.handlerId,
            createdAt = record.createdAt,
        )

    private fun createFailureContext(
        record: OutboxRecord<*>,
        retriesExhausted: Boolean,
        nonRetryable: Boolean,
    ): OutboxFailureContext =
        OutboxFailureContext(
            recordId = record.id,
            failureCount = record.failureCount,
            lastFailureReason = record.failureReason,
            handlerId = record.handlerId,
            retriesExhausted = retriesExhausted,
            nonRetryableException = nonRetryable,
        )
}
```

```kotlin
class OutboxFallbackHandlerInvoker(
    private val fallbackHandlerRegistry: OutboxFallbackHandlerRegistry,
) {
    fun hasFallback(handlerId: String): Boolean = fallbackHandlerRegistry.getByHandlerId(handlerId) != null

    fun dispatch(
        payload: Any?,
        metadata: OutboxRecordMetadata,
        context: OutboxFailureContext,
    ) {
        if (payload == null) return

        val fallbackHandler =
            fallbackHandlerRegistry.getByHandlerId(metadata.handlerId)
                ?: throw IllegalStateException("No fallback handler registered for handlerId: ${metadata.handlerId}")

        fallbackHandler.invoke(payload, metadata, context)
    }
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/fallback/factory/TypedFallbackHandlerMethodFactory.kt
```diff
@@ -0,0 +1,42 @@
...
+ * @author Roland Beisel
+ * @since 0.5.0
+ */
+class TypedFallbackHandlerMethodFactory : OutboxFallbackHandlerMethodFactory {
+    /**
+     * Checks if method matches typed fallback signature (3 params, first NOT Any).
+     */
+    override fun supports(method: Method): Boolean {
+        if (method.parameterCount != 3) return false
+        return method.parameterTypes.first().kotlin != Any::class
```
Alek96: I think we should be more strict here and we should check all parameters:
```kotlin
override fun supports(method: Method): Boolean {
    if (method.parameterCount != 3) return false

    val payloadType = method.parameterTypes[0].kotlin
    val metadataType = method.parameterTypes[1].kotlin
    val failureContext = method.parameterTypes[2].kotlin

    return payloadType != Any::class &&
        metadataType == OutboxRecordMetadata::class &&
        failureContext == OutboxFailureContext::class
}
``

File: namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/scanner/handler/AnnotatedHandlerScanner.kt [hunk 6]
```diff
@@ -0,0 +1,208 @@
...
+     *    b. Find matching @OutboxFallbackHandler via payload type
+     *    c. Create fallback wrapper if found
+     *    d. Return HandlerScanResult with handler + optional fallback
+     *
+     * @param bean The bean to scan for @OutboxHandler methods
+     * @return List of discovered handler scan results (empty if none found)
+     */
+    override fun scan(bean: Any): List<HandlerScanResult> {
+        // Find all fallback methods once (reused for all handlers)
+        val fallbackMethods = ReflectionUtils.findAnnotatedMethods(bean, OutboxFallbackHandler::class.java).toList()
```
Alek96: We could return here only fallback methods that make sense, not all methods annotated with `OutboxFallbackHandler`.
```kotlin
private fun getFallbackMethods(bean: Any): List<Method> =
    ReflectionUtils.findAnnotatedMethods(bean, OutboxFallbackHandler::class.java)
        .filter { handlerFactories.any { factory -> factory.supports(it) } }
        .toList()
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/scanner/handler/AnnotatedHandlerScanner.kt
```diff
@@ -0,0 +1,208 @@
...
+     * @return Fallback handler method or null if no match found
+     */
+    private fun findMatchingFallback(
+        bean: Any,
+        handlerMethod: Method,
+        fallbackMethods: List<Method>,
+    ): OutboxFallbackHandlerMethod? {
+        val handlerPayloadType = handlerMethod.parameterTypes.firstOrNull() ?: return null
+
+        // 1. Try exact type match first (typed fallback for typed handler)
```
Alek96: I am not sure what to think about looking for a typed fallback with the same payload type first and then, if not found, looking for a fallback with `Any` payload type. On the one hand, we have a fallback method for picking a fallback method - using a concrete fallback method or a general fallback one. On the other hand, this could introduce silent bugs. A user could, by mistake, put both fallback methods (Typed fallback and General) in the same class, and be surprised that only one of them is executed, but not the other (no warning that the General fallback was not selected).

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/scanner/handler/AnnotatedHandlerScanner.kt
```diff
@@ -0,0 +1,208 @@
...
+    private fun createFallbackFromMatch(
+        bean: Any,
+        matchingMethods: List<Method>,
+        handlerPayloadType: Class<*>,
+        matchType: String,
+    ): OutboxFallbackHandlerMethod? {
+        // Warn if multiple fallbacks found
+        if (matchingMethods.size > 1) {
+            log.warn(
+                "Multiple {} fallback handlers found for payload type {} in {}. Using first: {}. Others: {}",
```
Alek96: I think we should be more strict here. Either a user provides only one fallback method that matches the handler, or we will throw an error. Without the error, we could introduce a silent bug - the user would annotate two methods with the same type, and be surprised why one time one method is invoked, and another time the second method is invoked

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/scanner/handler/AnnotatedHandlerScanner.kt
```diff
@@ -0,0 +1,208 @@
...
+ * 2. **Generic fallback (priority 2)**: Fallback with Any payload type (catches all)
+ * 3. **First match wins**: If multiple fallbacks of same priority exist, first is used
+ * 4. **Warning on duplicates**: Logs warning if multiple fallbacks found
+ *
+ * This allows both specific typed fallbacks and generic catch-all fallbacks to coexist.
+ *
+ * @author Roland Beisel
+ * @since 0.4.0
+ */
+class AnnotatedHandlerScanner : HandlerScanner {
```
Alek96, line 25: I created a code that summarizes all comments below:
```kotlin
class AnnotatedHandlerScanner : HandlerScanner {
    private val log = LoggerFactory.getLogger(AnnotatedHandlerScanner::class.java)

    private val handlerFactories: List<OutboxHandlerMethodFactory> =
        listOf(
            TypedHandlerMethodFactory(),
            GenericHandlerMethodFactory(),
        )

    private val fallbackFactories =
        listOf(
            TypedFallbackHandlerMethodFactory(),
            GenericFallbackHandlerMethodFactory(),
        )

    override fun scan(bean: Any): List<HandlerScanResult> {
        // Find all fallback methods once (reused for all handlers)
        val fallbackMethods = getFallbackMethods(bean)

        return ReflectionUtils
            .findAnnotatedMethods(bean, OutboxHandler::class.java)
            .mapNotNull { handlerMethod ->
                // Create handler
                val handler = createHandler(bean, handlerMethod) ?: return@mapNotNull null

                // Find matching fallback based on payload type
                val fallback = findMatchingFallback(bean, handlerMethod, fallbackMethods)

                HandlerScanResult(handler, fallback)
            }.toList()
    }

    private fun getFallbackMethods(bean: Any): List<Method> =
        ReflectionUtils.findAnnotatedMethods(bean, OutboxFallbackHandler::class.java)
            .filter { handlerFactories.any { factory -> factory.supports(it) } }
            .toList()

    private fun createHandler(
        bean: Any,
        handlerMethod: Method,
    ): OutboxHandlerMethod? {
        val handlerFactory = handlerFactories.firstOrNull { factory -> factory.supports(handlerMethod) }
        if (handlerFactory == null) {
            log.warn(
                "No factory supports handler method {} in {}",
                handlerMethod.name,
                bean::class.simpleName,
            )
            return null
        }

        return handlerFactory.create(bean, handlerMethod)
    }

    private fun findMatchingFallback(
        bean: Any,
        handlerMethod: Method,
        fallbackMethods: List<Method>,
    ): OutboxFallbackHandlerMethod? {
        val handlerPayloadType = handlerMethod.parameterTypes.firstOrNull() ?: return null

        val matches =
            fallbackMethods
                .filter {
                    it.parameterTypes[0] == handlerPayloadType || it.parameterTypes[0] == Any::class.java
                }

        return createFallbackFromMatch(bean, matches, handlerPayloadType)
    }

    private fun createFallbackFromMatch(
        bean: Any,
        matchingMethods: List<Method>,
        handlerPayloadType: Class<*>,
    ): OutboxFallbackHandlerMethod? {
        if (matchingMethods.isEmpty()) return null

        check(matchingMethods.size <= 1) {
            "Multiple fallback handlers found for payload type ${handlerPayloadType.simpleName} in ${bean::class.simpleName}: " + matchingMethods.joinToString { it.name }
        }

        val fallbackMethod = matchingMethods.first()

        // Find factory and create fallback handler
        return fallbackFactories
            .firstOrNull { it.supports(fallbackMethod) }
            ?.create(bean, fallbackMethod)
    }
}

```

rolandbeisel: We have to use fallbackFactories in getFallbackMethods(). I removed the support for generic fallback handler registration for typed handlers to be more consistent. Furthermore I don't think that we should throw an exception if multiple matching fallback handlers were found (createFallbackFromMatch). Instead we should log (WARN), that only the first one is being registered.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/scanner/handler/AnnotatedHandlerScanner.kt
rolandbeisel: I removed the generic fallback to be more consistent

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
```diff
@@ -126,253 +68,89 @@ class OutboxProcessingScheduler(
...
-                    "Deleting outbox record {} after successful processing (deleteCompletedRecords=true)",
-                    record.id,
-                )
-                recordRepository.deleteById(record.id)
-            } else {
-                log.trace("Marking outbox record {} as COMPLETED", record.id)
-                record.markCompleted(clock)
-                recordRepository.save(record)
+            for (record in records) {
+                val shouldStop = processRecord(record, recordKey)
```
Alek96: Consider changing the `processRecord` return value to indicate whether processing should continue. While the current implementation works, it might be more intuitive if the method returned `true` when execution completes successfully and `false` when an error occurs.

After the change, we could have something like that:
```kotlin
private fun processRecord(
    record: OutboxRecord<*>,
    recordKey: String,
): Boolean {
    if (!record.canBeRetried(clock)) {
        log.trace("Skipping record {} - not ready for retry", record.id)
        return shouldContinueProcessing(recordKey)
    }

    return recordProcessorChain.handle(record)
        || shouldContinueProcessing(recordKey)
}

private fun shouldContinueProcessing(recordKey: String): Boolean {
    if (properties.processing.stopOnFirstFailure) {
        log.trace("Stopping key {} processing (stopOnFirstFailure=true)", recordKey)
        return false
    }

    return true
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxCoreAutoConfiguration.kt
```diff
@@ -264,73 +265,98 @@ class OutboxCoreAutoConfiguration {
...
+            FallbackOutboxRecordProcessor(
+                recordRepository = recordRepository,
+                fallbackHandlerInvoker = fallbackHandlerInvoker,
+                retryPolicyRegistry = retryPolicyRegistry,
+                properties = properties,
+                clock = clock,
+            )
+        val permanentFailure = PermanentFailureOutboxRecordProcessor(recordRepository)
+
+        return primary.apply {
```
Alek96: Proposition: If `setNext` returned the configured `OutboxRecordProcessor`, we could write code like this:
```kotlin
primary
    .setNext(retry)
    .setNext(fallback)
    .setNext(permanentFailure)
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/processor/RetryOutboxRecordProcessor.kt
```diff
@@ -0,0 +1,72 @@
...
+    private val retryPolicyRegistry: OutboxRetryPolicyRegistry,
+    private val recordRepository: OutboxRecordRepository,
+    private val clock: Clock,
+) : OutboxRecordProcessor() {
+    private val log = LoggerFactory.getLogger(RetryOutboxRecordProcessor::class.java)
+
+    /**
+     * Processes record by scheduling retry if possible.
+     *
+     * @return true if retry was scheduled, false if next processor should process
```
Alek96: The comment is incorrect. We should return `true` if the record was completed; otherwise, return `false`.
We should also change the base class comment.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/processor/PermanentFailureOutboxRecordProcessor.kt
```diff
@@ -0,0 +1,41 @@
...
+        recordRepository.save(record)
+
+        log.warn(
+            "Record {} for key {} marked as FAILED permanently after {} failures",
+            record.id,
+            record.key,
+            record.failureCount,
+        )
+
+        return handleNext(record)
```
Alek96: Since this is the final step in the processing chain and, by design, there should be no subsequent step, it might be better to return `false` explicitly rather than relying on the default behavior of `handleNext(record)`. To ensure that no other class violates the rule that this must be the last step in the chain, we could also override the `setNext` method:
```kotlin
override fun setNext(nextProcessor: OutboxRecordProcessor): OutboxRecordProcessor {
    throw UnsupportedOperationException("PermanentFailureOutboxRecordProcessor cannot have a next processor")
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/processor/FallbackOutboxRecordProcessor.kt
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @return true if fallback handler succeeded, false if next processor should process
+     */
+    override fun handle(record: OutboxRecord<*>): Boolean {
+        try {
+            val failureException = getFailureException(record)
+
+            log.debug("Dispatching record {} to fallback handler", record.id)
+            val success =
+                fallbackHandlerInvoker.dispatch(
```
Alek96, line 46: The `success` result adds some unnecessary complexity. Instead, we could first check whether a fallback method exists. If it doesn’t, we call `handleNext(record)`. If it does, we invoke the fallback method.
Sample code:
```kotlin
override fun handle(record: OutboxRecord<*>): Boolean {
    try {
        if (fallbackHandlerInvoker.hasFallback(record.handlerId)) {
            log.debug("Dispatching record {} to fallback handler", record.id)
            fallbackHandlerInvoker.dispatch(
                payload = record.payload,
                context = record.toFailureContext(getFailureException(record), retryPolicyRegistry),
            )

            completeRecord(record, recordRepository, properties, clock)
            return true
        }
        
        return handleNext(record)
    } catch (ex: Exception) {
        log.error("Fallback handler failed for record {}: {}", record.id, ex.message, ex)

        record.updateFailureException(ex)

        return handleNext(record)
    }
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/scanner/handler/AnnotatedHandlerScanner.kt
```diff
@@ -0,0 +1,124 @@
...
+     * Creates fallback handler from first matching method. Validates uniqueness.
+     */
+    private fun createFallbackFromMatch(
+        bean: Any,
+        matchingMethods: List<Method>,
+        handlerPayloadType: Class<*>,
+    ): OutboxFallbackHandlerMethod? {
+        if (matchingMethods.isEmpty()) return null
+
+        if (matchingMethods.size > 1) {
```
Alek96, line 110: My comment:
> I think we should be more strict here. Either a user provides only one fallback method that matches the handler, or we will throw an error. Without the error, we could introduce a silent bug - the user would annotate two methods with the same type, and be surprised why one time one method is invoked, and another time the second method is invoked

Your answer
>  I don't think that we should throw an exception if multiple matching fallback handlers were found (createFallbackFromMatch). Instead we should log (WARN), that only the first one is being registered.

One last concern: `Class.getMethods()` does not guarantee any particular order, so we shouldn’t rely on it returning methods in a consistent order across different JVM versions or executions. This could lead to a subtle bug where the application picks one annotated method in a local environment and a different one in production.

That said, I’ll defer to your decision here 😉

<br>
<br>

**Class getMethods doc:**
<p> The elements in the returned array are not sorted and are not in any
particular order.

<p> Generally, the result is computed as with the following 4 step algorithm.
Let C be the class or interface represented by this {@code Class} object:
<ol>
<li> A union of methods is composed of:
  <ol type="a">
  <li> C's declared public instance and static methods as returned by
       {@link #getDeclaredMethods()} and filtered to include only public
       methods.</li>
  <li> If C is a class other than {@code Object}, then include the result
       of invoking this algorithm recursively on the superclass of C.</li>
  <li> Include the results of invoking this algorithm recursively on all
       direct superinterfaces of C, but include only instance methods.</li>
  </ol></li>
<li> Union from step 1 is partitioned into subsets of methods with same
     signature (name, parameter types) and return type.</li>
<li> Within each such subset only the most specific methods are selected.
     Let method M be a method from a set of methods with same signature
     and return type. M is most specific if there is no such method
     N != M from the same set, such that N is more specific than M.
     N is more specific than M if:
  <ol type="a">
  <li> N is declared by a class and M is declared by an interface; or</li>
  <li> N and M are both declared by classes or both by interfaces and
       N's declaring type is the same as or a subtype of M's declaring type
       (clearly, if M's and N's declaring types are the same type, then
       M and N are the same method).</li>
  </ol></li>
<li> The result of this algorithm is the union of all selected methods from
     step 3.</li>
</ol>

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/processor/PermanentFailureOutboxRecordProcessor.kt
rolandbeisel: I do not aggree here, because each individual step should not know anthing about the overall chain. It cannot know, that it's the last one in the chain.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/scanner/handler/AnnotatedHandlerScanner.kt
rolandbeisel: Yes, correct, but if the user implements multiple fallback methods, then he introduced the bug, not the Lib. It will result in unexpected behaviour. This should be made clear in the docs.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/processor/FallbackOutboxRecordProcessor.kt
rolandbeisel: The suggested change would introduce a check-then-act pattern that we should avoid here for several reasons:

- The payload type is generic (OutboxRecord<T>), which means users can create records with nullable types like OutboxRecord<String?> (even if it doesn't make sense). The fallbackHandlerInvoker.dispatch() already handles this case internally and returns false when the payload is null.
- The boolean return value from dispatch() communicates important information:
false: No fallback handler registered OR payload is null
true: Fallback handler was successfully invoked

Both false cases require passing control to the next processor in the chain, which is exactly what we need.

If we called completeRecord() immediately after checking hasFallback(), we'd mark the record as completed even when dispatch() returns false (e.g., null payload). This would lead to incorrect behavior.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/processor/PermanentFailureOutboxRecordProcessor.kt
Alek96: I agree that each step cannot know if there exists a next step, but the step can stop processing at any time, and does not have to call `handleNext(record)`.
PrimaryOutboxRecordProcessor:
```kotlin
class PrimaryOutboxRecordProcessor() : OutboxRecordProcessor() {
    private val log = LoggerFactory.getLogger(PrimaryOutboxRecordProcessor::class.java)

    override fun handle(record: OutboxRecord<*>): Boolean {
        try {
            log.trace("Dispatching record {} to handler {}", record.id, record.handlerId)
            handlerInvoker.dispatch(record.payload, record.toMetadata())

            completeRecord(record, recordRepository, properties, clock)

            return true // <-------------------------------
        } catch (ex: Exception) {
            log.debug("Handler failed for record {} (key: {}): {}", record.id, record.key, ex.message)

            record.incrementFailureCount()
            record.updateFailureException(ex)

            return handleNext(record)
        }
    }
}
```
RetryOutboxRecordProcessor:
```kotlin
class RetryOutboxRecordProcessor : OutboxRecordProcessor() {

    override fun handle(record: OutboxRecord<*>): Boolean {
        val retryPolicy = retryPolicyRegistry.getByHandlerId(record.handlerId)

        if (shouldRetry(record, retryPolicy)) {
            val delay = retryPolicy.nextDelay(record.failureCount)
            record.scheduleNextRetry(delay, clock)
            recordRepository.save(record)

            log.debug("Scheduled retry #{} for record {} in {}", record.failureCount, record.id, delay)

            return false // <-----------------------------------------
        }

        return handleNext(record)
    }
```
FallbackOutboxRecordProcessor
```kotlin
class FallbackOutboxRecordProcessor() : OutboxRecordProcessor() {

    override fun handle(record: OutboxRecord<*>): Boolean {
        try {
            if (fallbackHandlerInvoker.hasFallback(record.handlerId)) {
                log.debug("Dispatching record {} to fallback handler", record.id)
                fallbackHandlerInvoker.dispatch(
                    payload = record.payload,
                    context = record.toFailureContext(getFailureException(record), retryPolicyRegistry),
                )

                completeRecord(record, recordRepository, properties, clock)
                return true // <-----------------------------------------
            }

            return handleNext(record)
        } catch (ex: Exception) {
            log.error("Fallback handler failed for record {}: {}", record.id, ex.message, ex)

            record.updateFailureException(ex)

            return handleNext(record)
        }
    }
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/processor/FallbackOutboxRecordProcessor.kt
Alek96: I agree that this solution is not ideal and has several problems, but we already have both of these problems present in the previous step - `PrimaryOutboxRecordProcessor`.
```kotlin
class PrimaryOutboxRecordProcessor(
    private val handlerInvoker: OutboxHandlerInvoker,
    private val recordRepository: OutboxRecordRepository,
    private val properties: OutboxProperties,
    private val clock: Clock,
) : OutboxRecordProcessor() {
    private val log = LoggerFactory.getLogger(PrimaryOutboxRecordProcessor::class.java)

    override fun handle(record: OutboxRecord<*>): Boolean {
        try {
            log.trace("Dispatching record {} to handler {}", record.id, record.handlerId)
            handlerInvoker.dispatch(record.payload, record.toMetadata())

            completeRecord(record, recordRepository, properties, clock)

            return true
        } catch (ex: Exception) {
            log.debug("Handler failed for record {} (key: {}): {}", record.id, record.key, ex.message)

            record.incrementFailureCount()
            record.updateFailureException(ex)

            return handleNext(record)
        }
    }
}

class OutboxHandlerInvoker(
    private val handlerRegistry: OutboxHandlerRegistry,
) {
    fun dispatch(
        payload: Any?,
        metadata: OutboxRecordMetadata,
    ) {
        if (payload == null) return

        val handler =
            handlerRegistry.getHandlerById(metadata.handlerId)
                ?: throw IllegalStateException("No handler with id ${metadata.handlerId}")

        when (handler) {
            is TypedHandlerMethod -> handler.invoke(payload)
            is GenericHandlerMethod -> handler.invoke(payload, metadata)
        }
    }
}
```

Alek96: That said, this comment is merely a suggestion

rolandbeisel: The key difference is: A handler must exist for every record
If no handler is found → IllegalStateException is thrown

This is correct behavior - missing handler is a configuration error

Fallback Handler: A fallback handler is optional
If no fallback is found → return false and pass to next processor
This is correct behavior - missing fallback is not an error

Your observation is correct: The OutboxHandlerInvoker doesn't need to return a boolean because:

- Handler is mandatory (exception if not found)
- Payload null check is just an early return (edge case for nullable types)
- No information needs to be communicated back

The OutboxFallbackHandlerInvoker needs the boolean because:

- Fallback is optional (false = no fallback found, continue processing)
- Payload can be null (false = can't process, continue to next processor)

Both cases require different handling than "handler not found"

The inconsistency you noticed exists because handlers and fallback handlers have fundamentally different semantics - one is mandatory, the other is optional.

Alek96: I was convinced :+1:

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/processor/PermanentFailureOutboxRecordProcessor.kt
rolandbeisel: OK, I implemented now a early return instead of calling `handleNext(record)`

Issue #115, opened 2025-12-15
Context Metadata Support for Outbox Records


rolandbeisel:
Add support for storing and retrieving arbitrary context metadata (e.g., tracing IDs, tenant IDs, correlation IDs) with outbox records.

New context: Map<String, String> field in OutboxRecord
Persisted as JSON in database (new context TEXT column)
Automatically serialized/deserialized by JPA mapper

**Use Cases**

- Distributed Tracing: Propagate trace/span IDs across async boundaries
- Multi-Tenancy: Include tenant ID for proper isolation in handlers
- Correlation IDs: Track requests across multiple services
- User Context: Store user ID for audit/security purposes
- Custom Metadata: Any domain-specific metadata needed during processing

**Benefits**

- Flexible: Support both global (provider) and local (per-record) context
- Non-Breaking: All changes are additive, existing code continues to work
- Type-Safe: Strongly typed interfaces with full KDoc documentation
- Performance: Provider context is collected once and cached
- Consistent: Works with both outbox.schedule() and @OutboxEvent annotation

rolandbeisel:
You can now safe context specific information (like traceparent) by using `OutboxContextProvider` Interface. This will be available with the next release. There is also a new example project, which shows, how to store and restore the tracing context.

PR #129, opened 2025-12-29
GH-115 Context Propagation

rolandbeisel:


gradle/libs.versions.toml
```diff
@@ -11,6 +11,7 @@ publish = "0.35.0"
 hibernate = "7.2.0.Final"
 springOrm = "7.0.2"
 springTx = "7.0.2"
+springAspects = "7.0.2"
```
Alek96: `springOrm`, `springTx`, and `springAspects` are all the `spring` version. We can consider merging them into one version.

namastack-outbox-tracing/build.gradle.kts
```diff
@@ -0,0 +1,22 @@
...
+}
+
+description = "namastack-outbox-tracing"
+
+dependencies {
+    implementation(project(":namastack-outbox-api"))
+    implementation(project(":namastack-outbox-core"))
+    implementation(libs.spring.boot.autoconfigure)
+    implementation(libs.spring.aspects)
+    implementation(libs.spring.boot.micrometer.tracing.opentelemetry)
```
Alek96: The tracing library should be based on the `micrometer-tracing` library, not on `opentelemetry`. Spring Boot uses Micrometer as its base, allowing users to choose which tracer library they prefer: OpenTelemetry, or OpenZipkin Brave. By supporting `micrometer-tracing`, we will support both.

https://docs.spring.io/spring-boot/reference/actuator/tracing.html

namastack-outbox-examples/namastack-outbox-example-tracing/build.gradle.kts
Alek96, line 1: I think we should merge the `micrometer` and `tracing` examples (for example, by renaming 'micrometer' to `tracing`, or by deleting `micrometer`). I created the `micrometer` example to show how to use trace with the library.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/handler/TypedHandlerMethod.kt
```diff
@@ -20,10 +21,16 @@ class TypedHandlerMethod(
...
      * @param payload Record payload matching paramType
+     * @param metadata Record metadata
      * @throws Throwable Original exception from handler (triggers retry logic)
      */
-    fun invoke(payload: Any) = invokeMethod(payload)
+    fun invoke(
+        payload: Any,
+        metadata: OutboxRecordMetadata,
+    ) {
+        invokeMethod(payload, metadata)
```
Alek96: I am unsure what your plan is for the `OutboxHandler` annotation, so the comment may be unnecessary. The metadata parameter in annotation-based handlers should be optional, allowing users to decide whether they want to include this argument or not.
```kotlin
@OutboxHandler
fun handle(payload: OrderCreatedEvent) {
    ...
}

@OutboxHandler
fun handle(payload: OrderCreatedEvent, matadata: OutboxRecordMetadata) {
    ...
}
```

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxSpanFactory.kt
```diff
@@ -1,84 +1,109 @@
...
+ *
+ * @param tracer Micrometer tracer for creating and managing spans
+ * @param propagator Micrometer propagator for extracting trace context from records
+ * @param handlerRegistry Registry for looking up handler metadata by handler ID
  */
 class OutboxSpanFactory(
     private val tracer: Tracer,
-    private val propagator: TextMapPropagator,
+    private val propagator: Propagator,
+    private val handlerRegistry: OutboxHandlerRegistry,
```
rolandbeisel: I don't think that's a good idea to use this internal core class here... We should rely on attributes, we have in the record. If there's something missing, we should add it there.

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxSpanFactory.kt
```diff
@@ -1,84 +1,109 @@
...
-                record.context,
-                OutboxTextMapGetter,
-            )
+            spanBuilder
+                .name(SPAN_NAME)
+                .kind(Span.Kind.PRODUCER)
+                .tag(ID_TAG, record.id)
+                .tag(KEY_TAG, record.key)
+                .tag(HANDLER_CLASS_TAG, handler.bean::class.java.name)
+                .tag(HANDLER_METHOD_TAG, handler.method.name)
```
rolandbeisel: tbh I don't get that... If you need the handler class and method, annotate your handler with @Observed and you're good to go. Spring does that for you

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxSpanFactory.kt
```diff
@@ -1,84 +1,109 @@
...
+                .tag(DELIVERY_ATTEMPT_TAG, record.failureCount + 1L)
```
rolandbeisel: In general I would provide the information we have from our record metadata and record itself here (as it was). It's not that much...

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxHandlerTracingAspect.kt
```diff
@@ -40,35 +43,36 @@ class OutboxHandlerTracingAspect(
...
-            span.recordException(ex)
-            span.setStatus(ERROR)
-
-            throw ex
+            return tracer.withSpan(span).use { block() }
         } finally {
+            record.failureException?.let { error -> span.error(error) }
             span.end()
         }
+    }
```
rolandbeisel: This is personal taste, but I like the old version much more, because it is very explicit:

1. Run Code in Span
2. Catch all exceptions
3. Set Span Status and Exception
4. Rethrow Exception
5. Finally end the span

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxHandlerTracingAspect.kt
```diff
@@ -40,35 +43,36 @@ class OutboxHandlerTracingAspect(
         joinPoint: ProceedingJoinPoint,
         record: OutboxRecord<*>,
     ): Any? {
-        val span = spanFactory.create(record) ?: return joinPoint.proceed()
+        val spanBuilder = spanFactory.create(record) ?: return joinPoint.proceed()
```
rolandbeisel: This looks weired now... A spanFactory, that creates a builder. And yes I know... the span was started in the factory before. I guess this is what you disliked. I think a SpanFactory should return a Span... and its created by calling `start()`

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxSpanFactory.kt
Alek96: My idea was to expose information about the handler itself—specifically, which class and which method are being invoked. I initially wanted to avoid using `handlerId`, since it is an internal identifier and could change in the future.

In theory, I could extract the same information from `handlerId` by parsing it (for example, using a regex like: `^(.+?)#([^(]+)\((.*)\)$`). However, on second thought, using `handlerId` may not be such a bad idea after all.

That said, having two explicit attributes—`class` and `method`—still feels more readable and self-explanatory.

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxSpanFactory.kt
Alek96: In my opinion, a *delivery attempt* is more intuitive and readable. When I see a span attribute like `delivery.attempt = 1`, it’s immediately clear that this is the first try. In contrast, if I see `failure.count = 1`, I need a moment to infer that this actually corresponds to the second attempt.

It also becomes ambiguous in failure scenarios. For example, if the second attempt fails, should `failure.count` still be `1`, or should it be `2`? This makes the semantics less clear, whereas `delivery.attempt` directly communicates what is happening without additional interpretation.

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxHandlerTracingAspect.kt
Alek96: I like your scenario as well, and I would use it. Unfortunately, our `outboxRecordProcessorChain.handle` method will never propagate a user exception. The processing chain always catches user errors and persists them in the record instead of throwing them.

With that in mind, should we record an exception if the fallback was completed successfully?

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxHandlerTracingAspect.kt
Alek96: I will change it so that the factory returns the Scope

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxSpanFactory.kt
rolandbeisel: My Problem with current implementation:

- Architectural Violation: The tracing module now depends on OutboxHandlerRegistry from core, creating a coupling that shouldn't exist. Tracing should be a lightweight add-on, not tightly coupled to core internals.

- Redundancy with User's `@Observed`: If users annotate their handler methods with `@Observed` (which is the Spring Boot way), they automatically get `code.function` and `code.namespace` tags on the handler's own span. The current implementation duplicates this information on the parent span.

- Wrong Level of Abstraction: The outbox process span represents the framework's processing, not the user's handler execution. Handler class/method information belongs on the handler's span (created by user's `@Observed`), not on the processing span.

For class/method information, users should add `@Observed` to their handler methods:
```kotlin
@OutboxHandler
@Observed
fun handleMyPayload(payload: MyPayload, metadata: OutboxRecordMetadata) {
...
}
```

This creates a proper span hierarchy:

- Parent: outbox process
- Child: User's handler method (with `code.namespace` and `code.function` from `@Observed`)

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxSpanFactory.kt
rolandbeisel: I'm convinced.

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxSpanFactory.kt
rolandbeisel: We're discussing this already in another conversation

namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/OutboxHandlerTracingAspect.kt
rolandbeisel: You're right in terms of throwing expcetions in the processing chain. These are being catched and not rethrowed. Thanks for pointing this out!

Nethertheless, I think that the following implementation is correct.

```kotlin
private inline fun <T> runWithSpan(
    span: Span,
    block: () -> T,
): T =
    try {
        tracer.withSpan(span).use { block() }
    } catch (ex: Exception) {
        span.error(ex)
        throw ex
    } finally {
        span.end()
    }
```

Let me explain why:

1. Framework vs. Business Separation: The span represents framework processing (try → retry → fallback → mark failed), not business outcomes. The framework succeeds even when business logic fails - this is working as designed.
2. Exception-Driven Error Model: The try-catch only marks errors when exceptions propagate out, indicating infrastructure failures (database crashes, configuration errors) rather than expected business failures handled through retries/fallbacks.
3. Clear and Safe: Activate span context, execute code, catch framework failures, always end span. This matches OpenTelemetry best practices and ensures spans are properly closed.
4. Correct Observability: Business failures are observable through outbox.record.failure_count tags, database status, and logs. Marking the framework span as "error" for expected business failures would create false alerts.
5. Messaging Semantics: In async messaging, consumer spans represent message delivery and framework processing, not business success. A properly handled failure (stored as FAILED) is successful framework operation.
6. User Handler Observability: When users annotate handlers with `@Observed`, those create child spans that automatically show errors on exceptions. This creates proper hierarchy: parent outbox process span shows framework success, child handler span shows business logic failure - clearly distinguishing infrastructure issues from business errors handled through retries/fallbacks.

PR #137, opened 2026-01-09
GH-89 add incremental outbox processing

Alek96:
I’ve migrated the changes from #97 with a few adjustments:

* Added an `ignoreRecordKeys` parameter to `findRecordKeysInPartitions` to ensure we always fetch up to `pollBatchSize` records:

  ```kotlin
  recordRepository.findRecordKeysInPartitions(
      partitions = assignedPartitions,
      status = NEW,
      batchSize = properties.processing.pollBatchSize,
      ignoreRecordKeys = unprocessedRecordKeys,      // <-----------
      ignoreRecordKeysWithPreviousFailure = properties.processing.stopOnFirstFailure,
  )
  ```

* Introduced the `properties.processing.concurrencyLimit` setting to cap the number of records processed concurrently. For example, a user can fetch records in batches of 10, while limiting the maximum number of concurrently processed records to 20.

* Reorganized several configuration properties:
  * `outbox.pollInterval` → `outbox.processing.pollInterval`
  * `outbox.batchSize` → `outbox.processing.pollBatchSize`
  * `outbox.rebalanceInterval` → `outbox.instance.rebalanceInterval`
  * `outbox.processing.publishAfterSave` → `outbox.multicaster.publishAfterSave`

### DIFF
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxEventMulticaster.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxEventMulticaster.kt
@@ -94,7 +94,7 @@ class OutboxEventMulticaster(
         log.debug("Saving @OutboxEvent to outbox: $classSimpleName")
         saveOutboxRecord(payload, annotation)
 
-        if (outboxProperties.processing.publishAfterSave) {
+        if (outboxProperties.multicaster.publishAfterSave) {
             log.debug("Publishing @OutboxEvent to listeners: $classSimpleName")
             delegateEventMulticaster.multicastEvent(event, eventType)
         }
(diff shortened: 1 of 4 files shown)


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
```diff
@@ -50,51 +49,71 @@ class OutboxProcessingScheduler(
      * Each record is processed through the processor chain which handles
      * dispatching, retry logic, fallback invocation, and failure marking.
      */
-    @Scheduled(fixedDelayString = "\${outbox.poll-interval:2000}", scheduler = "outboxDefaultScheduler")
+    @Scheduled(fixedDelayString = "\${outbox.processing.poll-interval:2000}", scheduler = "outboxDefaultScheduler")
```
rolandbeisel, line 52: changing poll-interval path would be a breaking change. we could support both and make one deprecated...

PR #139, opened 2026-01-10
JDBC Module

rolandbeisel:


namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/config/JdbcOutboxAutoConfiguration.kt
```diff
@@ -0,0 +1,145 @@
...
+
+    /**
+     * Creates a transaction template for outbox operations.
+     *
+     * @param transactionManager Platform transaction manager
+     * @return Transaction template for outbox operations
+     */
+    @Bean("outboxTransactionTemplate")
+    @ConditionalOnMissingBean(name = ["outboxTransactionTemplate"])
+    internal fun outboxTransactionTemplate(transactionManager: PlatformTransactionManager): TransactionTemplate =
```
Alek96, line 75: `TransactionTemplate` is autoconfigured by `org.springframework.transaction.support.TransactionTemplate.TransactionAutoConfiguration` and it is annotated with `@ConditionalOnMissingBean(TransactionOperations.class)`, which means that after our library creates `TransactionTemplate` bean, it can override the default Spring bean.

```java
@AutoConfiguration
@ConditionalOnClass(PlatformTransactionManager.class)
public final class TransactionAutoConfiguration {

	@Configuration(proxyBeanMethods = false)
	@ConditionalOnSingleCandidate(PlatformTransactionManager.class)
	static class TransactionTemplateConfiguration {

		@Bean
		@ConditionalOnMissingBean(TransactionOperations.class)
		TransactionTemplate transactionTemplate(PlatformTransactionManager transactionManager) {
			return new TransactionTemplate(transactionManager);
		}

	}
```

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxAutoConfiguration.kt
```diff
@@ -0,0 +1,145 @@
...
+     * Creates a named JdbcTemplate for outbox operations.
+     *
+     * Users can override this by providing their own "outboxJdbcTemplate" bean.
+     *
+     * @param dataSource DataSource to use
+     * @return JdbcTemplate for outbox operations
+     */
+    @Bean("outboxJdbcTemplate")
+    @ConditionalOnMissingBean(name = ["outboxJdbcTemplate"])
+    internal fun outboxJdbcTemplate(dataSource: DataSource): JdbcTemplate = JdbcTemplate(dataSource)
```
Alek96: I am not sure if I see many advantages of using a custom `JdbcTemplate` bean. I think we should use the default Spring `JdbcTemplate` bean.

The default Spring `JdbcTemplate` bean is created by `org.springframework.boot.jdbc.autoconfigure.JdbcTemplateAutoConfiguration`

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/config/JdbcOutboxAutoConfiguration.kt
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * Auto-configuration class for JDBC-based Outbox functionality.
+ *
+ * This configuration provides JDBC implementations for outbox repositories
+ * and handles database schema initialization when enabled.
+ *
+ * @author Roland Beisel
+ * @since 1.1.0
+ */
+@AutoConfiguration
```
Alek96, line 36: To be safe, we can add information that the `JdbcOutboxAutoConfiguration` should be run after `JdbcTemplateAutoConfiguration`.

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxRecordRepository.kt
```diff
@@ -0,0 +1,363 @@
...
+        GROUP BY o.record_key
+        ORDER BY MIN(o.created_at) ASC
+        LIMIT ?
+    """
+
+    /**
+     * Saves an outbox record.
+     * Updates existing record or inserts new one.
+     */
+    override fun <T> save(record: OutboxRecord<T>): OutboxRecord<T> =
```
Alek96, line 148: I do not like the idea of trying to update the record, and if it does not exist, we will insert it (the JPA has this limitation that many people dislike). Let's introduce two new methods - 'insert` and `update` - and use them instead of `save` in our code.

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxRecordRepository.kt
```diff
@@ -0,0 +1,363 @@
...
+        transactionTemplate.execute {
+            val payload = record.payload ?: throw IllegalArgumentException("record payload cannot be null")
```
Alek96: An idea to consider: create `OutboxRecordEntity` for JDBC and `OutboxRecordEntityMapper`. I know this is a lot of classes for a single method, but if you go with 'insert' and 'update` methods, then it will be more useful.

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxRecordRepository.kt
```diff
@@ -0,0 +1,363 @@
...
+    /**
+     * Attempts to update an existing record. Returns number of rows updated.
+     */
+    private fun <T> tryUpdate(
+        record: OutboxRecord<T>,
+        recordType: String,
+        serializedPayload: String,
+        serializedContext: String?,
+    ): Int =
+        jdbcTemplate.update(
```
Alek96: Consider using `JdcbClient` ([link](https://docs.spring.io/spring-boot/reference/data/sql.html#data.sql.jdbc-client)) instead of `jdbcTemplate`:
```kotlin
jdbcClient
    .sql(updateRecordQuery.trimIndent())
    .param("id", record.id)
    .param("status", record.status.name)
    .param("record_key", record.key)
    .param("record_type", recordType)
    .param("payload", serializedPayload)
    .param("context", serializedContext)
    .param("partition_no", record.partition)
    .param("created_at", record.createdAt)
    .param("completed_at", record.completedAt)
    .param("failure_count", record.failureCount)
    .param("failure_reason", record.failureReason)
    .param("next_retry_at", record.nextRetryAt)
    .param("handler_id", record.handlerId)
    .update();
```

namastack-outbox-jdbc/src/main/resources/schema/h2/outbox-tables.sql
Alek96, line 1: Is this file a duplicate of the file from the JPA module? I wonder if extracting schema initialization to an external module would be a good idea.

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxRecordRepository.kt
rolandbeisel: Love that... I was not aware of this.

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxRecordRepository.kt
rolandbeisel: I don't want to change the repository interface for this... and i also don't see a big problem here.

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxAutoConfiguration.kt
rolandbeisel: This pattern makes sense for multiple reasons:

1. Multiple DataSources: When you have multiple datasources in your application, you can create your own `outboxJdbcClient` bean that uses a specific datasource (e.g., a separate outbox database). Without the name qualifier, Spring wouldn't know which JdbcClient to inject.
2. Explicit Bean Name: The `@Bean("outboxJdbcClient")` gives it a specific name so it can be referenced via `@Qualifier("outboxJdbcClient")` in other beans (like in outboxRecordRepository).
3. Override Capability: The `@ConditionalOnMissingBean(name = ["outboxJdbcClient"])` allows users to provide their own custom `outboxJdbcClient` bean with specific configurations (connection pool settings, custom datasource, etc.) and the auto-configuration will respect that.
4. Isolation: It keeps the outbox's JDBC client separate from any other JdbcClient beans the application might have, preventing bean conflicts and ambiguity.

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/config/JdbcOutboxAutoConfiguration.kt
rolandbeisel: same reason like for outboxJdbcTemplate... e.g. multiple datasources

Alek96: The multiple data sources argument is valid. What I wanted to highlight is that once we configure a custom `TransactionTemplate` bean, we effectively override Spring’s default one. More precisely, we create our own bean instead of using the one provided by Spring.

In a scenario where the default datasource points to database A and the outbox datasource points to database B, the outbox `TransactionTemplate` would become the primary `TransactionTemplate` bean. As a result, the primary `TransactionTemplate` would point to database B instead of database A, which could lead to unexpected behavior.

namastack-outbox-jdbc/src/main/resources/schema/h2/outbox-tables.sql
rolandbeisel: Thought about this of course. The duplication here is actually acceptable architectural duplication because JPA and JDBC are independent alternatives. Forcing them to share schema infrastructure could create unwanted coupling.

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/config/JdbcOutboxAutoConfiguration.kt
rolandbeisel: You're right that Spring's default transactionTemplate bean is not created.

Single datasource scenario: When there's only one datasource, both the (non-existent) default transactionTemplate and outboxTransactionTemplate would have used the same PlatformTransactionManager anyway, so there's no functional difference.

Multiple datasources scenario: Users override the outboxTransactionTemplate bean to point to their specific datasource.

Yes, applications using this library won't have Spring's default transactionTemplate bean available. I think this is acceptable because:
- Modern Spring applications should use `@Transactional` annotations for declarative transaction management
- Direct TransactionTemplate injection is rarely needed in application code
- The explicit naming pattern is necessary to support the multi-datasource use case

The only impact: If users were directly injecting Spring's default transactionTemplate bean by name, that will fail. They need to create their own or use `@Transactional`.

If users run into trouble with this setup, we can change it. But multi-db support was in fact a user request.

rolandbeisel: Ah... we could add @AutoConfigureAfter(TransactionAutoConfiguration::class)... i think this would solve all problems for transactionTemplate

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxRecordRepository.kt
```diff
@@ -0,0 +1,348 @@
...
+                .param(id)
+                .update()
+        }
+    }
+
+    /**
+     * Finds record keys in specified partitions that are ready for processing.
+     * Optionally filters out keys with previous incomplete records.
+     */
+    override fun findRecordKeysInPartitions(
```
Alek96, line 257: Spring's `JdbcClient` with an `IN` clause works well with named parameters or positional parameters as they can access a list argument:
```kotlin
private val recordKeysQueryWithPreviousFailureFilterTemplate = """
    SELECT o.record_key
    FROM outbox_record o
    WHERE o.partition_no IN (:partition_nos)
      AND o.status = :status
      AND o.next_retry_at <= :next_retry_lte
      AND NOT EXISTS (
        SELECT 1 FROM outbox_record older
        WHERE older.record_key = o.record_key
          AND older.completed_at IS NULL
          AND older.created_at < o.created_at
      )
    GROUP BY o.record_key
    ORDER BY MIN(o.created_at) ASC
    LIMIT :limit
"""

private val recordKeysQueryWithoutPreviousFailureFilterTemplate = """
    SELECT o.record_key
    FROM outbox_record o
    WHERE o.partition_no IN (:partition_nos)
      AND o.status = :status
      AND o.next_retry_at <= :next_retry_lte
    GROUP BY o.record_key
    ORDER BY MIN(o.created_at) ASC
    LIMIT :limit
"""

override fun findRecordKeysInPartitions(
    partitions: Set<Int>,
    status: OutboxRecordStatus,
    batchSize: Int,
    ignoreRecordKeysWithPreviousFailure: Boolean,
): List<String> {
    val query = if (ignoreRecordKeysWithPreviousFailure) {
        recordKeysQueryWithPreviousFailureFilterTemplate
    } else {
        recordKeysQueryWithoutPreviousFailureFilterTemplate
    }

    return jdbcClient.sql(query)
        .param("partition_nos", partitions.toList())
        .param("status", status.name)
        .param("next_retry_lte", OffsetDateTime.now(clock))
        .param("limit", batchSize)
        .query(String::class.java)
        .list()
        .filterNotNull()
}
```

namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxInstanceRepository.kt
```diff
@@ -0,0 +1,278 @@
...
+
+    /**
+     * Finds all active instances.
+     */
+    override fun findActiveInstances(): List<OutboxInstance> = findByStatus(OutboxInstanceStatus.ACTIVE)
+
+    /**
+     * Finds instances with stale heartbeat.
+     */
+    override fun findInstancesWithStaleHeartbeat(cutoffTime: OffsetDateTime): List<OutboxInstance> {
```
Alek96, line 170: The same comment as for `findRecordKeysInPartitions` - Spring's `JdbcClient` with an `IN` clause works well with named parameters or positional parameters, as they can access a list argument:
```kotlin
private val findStaleHeartbeatQueryTemplate = """
    SELECT * FROM outbox_instance
    WHERE last_heartbeat < :last_heartbeat_lt
      AND status IN (:statues)
    ORDER BY last_heartbeat
"""

override fun findInstancesWithStaleHeartbeat(cutoffTime: OffsetDateTime): List<OutboxInstance> {
    val activeStatuses = listOf(OutboxInstanceStatus.ACTIVE, OutboxInstanceStatus.SHUTTING_DOWN)

    return jdbcClient
        .sql(findStaleHeartbeatQueryTemplate.trimIndent())
        .param("last_heartbeat_lt", cutoffTime)
        .param("statues", activeStatuses.map { it.name })
        .query(JdbcOutboxInstanceEntity::class.java)
        .list()
        .filterNotNull()
        .map { JdbcOutboxInstanceEntityMapper.map(it) }
}
```

Issue #50, opened 2025-10-27
Kafka Integration


rolandbeisel:
Ready-made processors for common brokers: Kafka

namastack-outbox-kafka

PR #191, opened 2026-02-06
GH-50 Add Kafka routing module with configuration and selectors

rolandbeisel:


namastack-outbox-kafka/src/main/kotlin/io/namastack/outbox/kafka/KafkaOutboxProperties.kt
```diff
@@ -0,0 +1,21 @@
...
+@ConfigurationProperties("namastack.outbox.kafka")
+data class KafkaOutboxProperties(
+    /**
+     * Whether Kafka outbox integration is enabled.
+     */
+    val enabled: Boolean = true,
+    /**
+     * Default Kafka topic for outbox events.
+     */
+    val defaultTopic: String = "outbox-events",
```
Alek96: If I understand the code correctly, a user must define a `KafkaRoutingConfiguration` bean whenever they want to deliver events to more than one queue. In practice, this means that in most cases a custom `KafkaRoutingConfiguration` bean will be provided, and the default configuration will no longer be used.

With that in mind, does it still make sense to have a property that, by default, will not be used?

namastack-outbox-kafka/src/main/kotlin/io/namastack/outbox/kafka/KafkaRoutingConfiguration.kt
```diff
@@ -0,0 +1,85 @@
...
+         * Creates a new Kafka routing configuration.
+         */
+        @JvmStatic
+        fun create(): KafkaRoutingConfiguration = KafkaRoutingConfiguration()
+    }
+
+    /**
+     * Configures routing rules with Kafka-specific DSL.
+     */
+    fun routing(configurer: KafkaRoutingConfigurer.() -> Unit): KafkaRoutingConfiguration {
```
Alek96: Is it possible for `create` to return a `KafkaRoutingConfigurer`, so that we could have an API like:

```kotlin
create()
  .route(...)
  .default(...)
```

I feel that having a class with a single `routing` method that always has to be invoked for the builder to work makes the API more complicated than necessary.

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/routing/OutboxRoutingRule.kt
```diff
@@ -0,0 +1,163 @@
...
+ * - [headers] - message headers/attributes
+ *
+ * Rules are evaluated in declaration order - first match wins.
+ *
+ * ## Example (Kotlin)
+ *
+ * ```kotlin
+ * OutboxRoutingRule.builder(OutboxPayloadSelector.type(OrderEvent::class.java))
+ *     .target("orders")
+ *     .key { payload, metadata -> (payload as OrderEvent).orderId }
```
Alek96: Idea - if we make Builder generic, then when using  `OutboxPayloadSelector.type(OrderEvent::class.java)` we could cast payload to the desired class for the user

namastack-outbox-kafka/src/main/kotlin/io/namastack/outbox/kafka/KafkaOutboxProperties.kt
rolandbeisel: I think there are many cases where you want to deliver all of your events to the same topic. Imagine a order microservice which has one primary aggregate root. All of its events would go to the same topic (topic orders, which contains OrderPlaced, OrderCanceled, ...)

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/routing/OutboxRoutingRule.kt
rolandbeisel: Good idea, but it doesn't work well.
While making the builder generic sounds appealing for type-safe access, it breaks down when using annotation or predicate selectors where multiple types can match the same route.

```kotin
route(OutboxPayloadSelector.annotation(Publishable::class.java)) {
    topic("events")
    key { payload, _ -> 
        // payload could be OrderEvent, PaymentEvent, or any @Publishable class
        // a generic T doesn't help here - we need to use a common interface
        (payload as Publishable).eventKey
    }
}
```

In this scenario, there's no single type T we could use. The user must handle the polymorphism themselves, either by casting to a common interface or using when expressions. This is actually the correct behavior - the routing DSL shouldn't pretend to know the concrete type when multiple types can match.

namastack-outbox-kafka/src/main/kotlin/io/namastack/outbox/kafka/KafkaRoutingConfiguration.kt
rolandbeisel: Optimized that: 

```kotlin
val config = kafkaRouting {
    route(OutboxPayloadSelector.type(OrderEvent::class.java)) {
        topic("orders")
        key { event, _ -> (event as OrderEvent).orderId }
        headers { _, metadata -> metadata.context }
        mapping { event, _ -> (event as OrderEvent).toPublicEvent() }
        filter { event, _ -> (event as OrderEvent).status != "CANCELLED" }
    }
    defaults {
        topic("domain-events")
    }
}
```

Issue #52, opened 2025-10-27
RabbitMQ Integration


rolandbeisel:
Ready-made processors for common brokers: RabbitMQ

namastack-outbox-rabbitmq

PR #194, opened 2026-02-09
GH-52 Add RabbitMQ integration and examples

rolandbeisel:


namastack-outbox-rabbit/src/main/kotlin/io/namastack/outbox/rabbit/RabbitOutboxProperties.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+/**
+ * Configuration properties for Rabbit outbox integration.
+ *
+ * @author Roland Beisel
+ * @since 1.1.0
+ */
+@ConfigurationProperties("namastack.outbox.rabbit")
+data class RabbitOutboxProperties(
```
Alek96, line 12: I would define `namastack.outbox.rabbit.enable-json` property here as well

Issue #223, opened 2026-03-08
Introduce Observability for Outbox Record Processing


Alek96:
## Overview
Enhance outbox record processing by introducing structured observability using Micrometer Observation and OpenTelemetry Semantic Conventions for messaging.

## Motivation
Currently, outbox processing lacks detailed observability, making it difficult to monitor, track, and troubleshoot the delivery lifecycle for each handler. A messaging-like model could improve traceability: when an event is scheduled, independent outbox records are created for each handler matching the event—similar to publishing an event on a topic for designated handlers.

## Proposal
Add support for process observability:
- Integrate Micrometer Observation ([docs](https://docs.micrometer.io/micrometer/reference/observation.html)) to instrument processing steps.
- Apply OpenTelemetry messaging semantic conventions ([docs](https://opentelemetry.io/docs/specs/semconv/messaging/)) for consistent trace data.
- Leverage context propagation as introduced in [GH-115](https://github.com/namastack/namastack-outbox/issues/115), ensuring observability data is correctly propagated through handler execution and delivery attempts.

### Observation Attributes
- name: `outbox.record.process`
- contextual name: `outbox process`

#### Low Cardinality Keys
- `outbox.handler.kind` — handler type (values: `primary`, `fallback`)
- `outbox.handler.id` — unique identifier: `class.name$methodName(arguments)`

#### High Cardinality Keys
- `outbox.record.id` — database record ID
- `outbox.record.key` — record key
- `outbox.delivery.attempt` — delivery attempt count

## Benefits
- Consistent and granular record-level tracing
- Improved troubleshooting and operational insight
- Alignment with industry standards

## Next Steps
- Instrument outbox processing logic
- Review handler creation and delivery mapping for trace context propagation
- Add documentation and examples

---
This proposal aims for a robust observability layer that works seamlessly with Micrometer and OpenTelemetry standards, leveraging context propagation from GH-115.

PR #225, opened 2026-03-09
Feature - Record processing observability

Alek96:
PR solves GH-223: Introduce Observability for Outbox Record Processing

### DIFF
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/invoker/OutboxFallbackHandlerInvoker.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/invoker/OutboxFallbackHandlerInvoker.kt
@@ -1,5 +1,6 @@
 package io.namastack.outbox.handler.invoker
 
+import io.namastack.outbox.OpenForProxy
 import io.namastack.outbox.OutboxRecord
 import io.namastack.outbox.handler.registry.OutboxFallbackHandlerRegistry
 import io.namastack.outbox.retry.OutboxRetryPolicyRegistry
@@ -15,7 +16,8 @@ import io.namastack.outbox.retry.OutboxRetryPolicyRegistry
  * @author Roland Beisel
  * @since 1.0.0
  */
-open class OutboxFallbackHandlerInvoker(
+@OpenForProxy
+class OutboxFallbackHandlerInvoker(
     private val retryPolicyRegistry: OutboxRetryPolicyRegistry,
     private val fallbackHandlerRegistry: OutboxFallbackHandlerRegistry,
 ) {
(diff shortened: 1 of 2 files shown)


namastack-outbox-tracing/src/main/kotlin/io/namastack/outbox/tracing/aop/OutboxInvokerObservationAdvice.kt
```diff
@@ -0,0 +1,72 @@
...
+        val args = invocation.arguments
+        val record = args.firstOrNull() as? OutboxRecord<*> ?: return invocation.proceed()
+
+        val observation =
+            OutboxObservationDocumentation.OUTBOX_RECORD_PROCESS.observation(
+                null,
+                OutboxObservationDocumentation.DefaultOutboxProcessObservationConvention.INSTANCE,
+                { OutboxProcessObservationContext(record, handlerKind) },
+                observationRegistry,
+            )
```
rolandbeisel, line 79: Should we support custom conventions here (first param)? Currently there is a DefaultOutboxProcessObservationConvention, but I'm wondering, if people want to use their own namings for example. This is non-blocking... I'm just interested what you think.

Alek96: From my understanding of the Micrometer documentation and the codebase, the correct approach is to create the observation using the default implementation and then customize it with `ObservationHandler` if needed.

If a user wants to change observation attributes — such as the name, contextual name, or adding/removing cardinality key-values — they can implement an `ObservationHandler` that supports only our custom `Observation.Context` and override those values in the `onStart` method.

Although `ObservationHandler` is primarily intended for reacting to observation lifecycle events and recording metrics, it can also be used to modify observation attributes.

Documentation: https://docs.micrometer.io/micrometer/reference/observation/components.html#micrometer-observation-handler

Example:
```java
class MyContext extends Observation.Context {

    MyContext() {
    }
}

class TenantEnrichingHandler implements ObservationHandler<MyContext> {

    @Override
    public void onStart(MyContext context) {
        context.addLowCardinalityKeyValue(KeyValue.of("tenant", SecurityUtils.getTenant()));
    }

    @Override
    public boolean supportsContext(Observation.Context context) {
        return context instanceof MyContext;
    }
}
```

But I am open to suggestions, as I am relatively new to Micrometer Observation and I could misunderstood it

rolandbeisel: I added support for custom conventions and tested it successfully

rolandbeisel (dismissed):

PR #228, opened 2026-03-10
feat(mongodb): Complete MongoDB Implementation with Transactional Safety and 100% Test Parity

Stellarhold170NT:
This PR significantly enhances the MongoDB support for namastack-outbox, bringing it to full feature and reliability parity with the existing JDBC/JPA implementations. The goal was not just to store records, but to ensure the same high standards for multi-instance high-availability and transactional integrity

With this update, users can switch from a relational database to MongoDB by simply swapping the starter dependency, without changing any business logic


namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/config/MongoOutboxAutoConfiguration.kt
```diff
@@ -0,0 +1,90 @@
...
+        mongoTemplate: MongoTemplate,
+        entityMapper: MongoOutboxRecordEntityMapper,
+    ): OutboxRecordRepository = MongoOutboxRecordRepository(mongoTemplate, entityMapper)
+
+    @Bean
+    @ConditionalOnMissingBean
+    internal fun outboxRecordStatusRepository(
+        mongoTemplate: MongoTemplate,
+        entityMapper: MongoOutboxRecordEntityMapper,
+    ): OutboxRecordStatusRepository = MongoOutboxRecordRepository(mongoTemplate, entityMapper)
```
rolandbeisel: Since MongoOutboxRecordRepository implements both OutboxRecordRepository and OutboxRecordStatusRepository, Spring automatically makes that single bean injectable by either interface type. The metrics module uses ObjectProvider<OutboxRecordStatusRepository> and successfully resolves it. I don't think that this is necessary.

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxRecordRepository.kt
```diff
@@ -0,0 +1,186 @@
...
+     * Race conditions can only occur during the brief rebalancing window when an
+     * instance fails.
+     */
+    override fun findRecordKeysInPartitions(
+        partitions: Set<Int>,
+        status: OutboxRecordStatus,
+        batchSize: Int,
+        ignoreRecordKeysWithPreviousFailure: Boolean
+    ): List<String> {
+        val now = Instant.now()
```
rolandbeisel: we need to inject a clock here.

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxRecordRepository.kt
```diff
@@ -0,0 +1,186 @@
...
+                Criteria.where("recordKey").`is`(record.recordKey)
+                    .and("completedAt").`is`(null)
+                    .and("createdAt").lt(record.createdAt)
+            ).limit(1)
+            !mongoTemplate.exists(olderIncompleteQuery, MongoOutboxRecordEntity::class.java)
+        }
+            .map { it.recordKey }
+            .distinct()
+            .take(batchSize)
+    }
```
rolandbeisel: This is quite interesting. So this logic over-fetches records and then filters them out. With a batch-size of 200 this would query 1000 records initially and would then call mongoTemplate.exists(...) for all of these records. I have to think about this. I think, I understood, why this is necessary.

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxRecordEntity.kt
```diff
@@ -0,0 +1,51 @@
...
+    @Indexed
+    val status: OutboxRecordStatus,
+    
+    val recordKey: String,
+    val recordType: String,
+    val payload: String,
+    val context: String?,
+    val partitionNo: Int,
+    
+    @Indexed
```
Alek96: I think you are not using this index

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxRecordEntity.kt
```diff
@@ -0,0 +1,51 @@
...
+    val createdAt: Instant,
+    
+    val completedAt: Instant?,
+    val failureCount: Int,
+    val failureReason: String?,
+    
+    @Indexed
```
Alek96: I think you are not using this index

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxInstanceEntity.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.data.mongodb.core.mapping.Document
+import java.time.Instant
+
+/**
+ * Entity representing an outbox instance in MongoDB.
+ *
+ * @author Stellar Hold
+ * @since 1.1.0
+ */
+@Document(collection = "outbox_instances")
```
Alek96, line 16: I think you are missing an index on `status`, `last_heartbeat`

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxPartitionAssignmentEntityMapper.kt
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * @author Stellar Hold
+ * @since 1.1.0
+ */
+internal object MongoOutboxPartitionAssignmentEntityMapper {
+    fun map(assignment: PartitionAssignment): MongoOutboxPartitionAssignmentEntity =
+        MongoOutboxPartitionAssignmentEntity(
+            partitionNumber = assignment.partitionNumber,
+            instanceId = assignment.instanceId,
+            version = assignment.version ?: 0,
```
Alek96: Would setting the version to `0` when it is `null` potentially cause an issue here?

Consider the scenario of a new database or the first application run:
* **Instance A** initializes all 256 partitions by inserting them with `version = 0`.
* **Instance B** detects that all 256 partitions already exist, so it attempts to update them using a `WHERE version = 0` condition. Since each partition entity also has `version = 0`, the updates would succeed.

In this case, both instances could successfully update the same partitions, which should not happen

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxRecordRepository.kt
Alek96: I asked LLM how to solve it, and this was its proposition: 

```kotlin
newAggregation(
    // 1. Get all uncompleted records (Assuming completedAt is null for incomplete ones)
    match(
        Criteria.where("completedAt").isNull
            .and("partitionNo").`in`(partitions)
    ),
    // 2. Sort by creation date
    sort(
        Sort.by(
            Sort.Order.asc("recordKey"),
            Sort.Order.asc("createdAt")
        )
    ),
    // 3. Group by recordKey, grabbing the first (oldest) document's properties
    group("recordKey")
        .first(Aggregation.ROOT).`as`("doc"),
    // 4. Ensure the oldest record actually matches our criteria to process NOW
    match(
        Criteria.where("doc.status").`is`(status.name)
            .and("doc.nextRetryAt").lte(now)
    ),
    // 5. Order by oldest overall
    sort(Sort.Direction.ASC, "doc.createdAt"),
    // 6. Limit to batch size
    limit(batchSize.toLong()),
    // 7. Project only the record_key (which is currently stored as _id from the group stage)
    project().andExclude("_id").and("_id").`as`("recordKey")
)
```

### Why this works perfectly for your Outbox logic:
If a `record_key` has an older task that failed or is not ready yet (e.g., its `next_retry_at` is in the future), the `$group` stage will grab that older task. When the pipeline moves to the second `$match` stage, that older task will fail the `next_retry_at <= :now` check, and the entire `record_key` will be filtered out. This guarantees strict FIFO (First-In-First-Out) processing per `record_key`!

### Indexing Tip for MongoDB:
To make this query fast, you should create a compound index that supports the initial filter and sort:
```
Index({ partitionNo: 1, completed_at: 1, record_key: 1, created_at: 1 })
```
or
```
Index(
  { partitionNo: 1, recordKey: 1, createdAt: 1 },
  { partialFilterExpression: { completedAt: null } }
)
```

Stellarhold170NT:
@rolandbeisel Hi, I just finished refactoring the Mongo module to match JDBC parity. I've also built a demo project (Spring Boot 4 + Mongo 8) here: https://github.com/Stellarhold170NT/namastack-outbox-sample-mongodb. Let me know what you think

tomwetjens:
Nice work! We have many services on MongoDB and we are really waiting for this to become released.

malczuuu:
Hi!

I have following questions:

1. For relational databases, this library provides a way to customize table prefixes (in jdbc module). How about doing something like that for mongodb as well?
2. For relational dbs, there's a migration script for setting up schemas. MongoDB entities in here use annotations for indexes, but not every project allows creation of indexes based on that. How about adding in example a reference mongosh/js script for setting up collections and indexes?

Kind regards,
Damian

Stellarhold170NT:
hi @malczuuu 
I will implement both features and update the PR as soon as possible. Thanks

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/config/MongoOutboxAutoConfiguration.kt
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Auto-configuration for MongoDB-based outbox implementation.
+ *
+ * @author Stellar Hold
+ * @since 1.1.0
+ */
+@AutoConfiguration
+@AutoConfigureBefore(OutboxCoreInfrastructureAutoConfiguration::class)
+@ConditionalOnClass(MongoTemplate::class)
+@ConditionalOnProperty(name = ["namastack.outbox.mongodb.enabled"], havingValue = "true", matchIfMissing = true)
```
rolandbeisel: why `namastack.outbox.mongodb.enabled`? It should be `namastack.outbox.enabled`

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/config/MongoOutboxAutoConfiguration.kt
```diff
@@ -0,0 +1,95 @@
...
+        TransactionTemplate(beanFactory.getBean<MongoTransactionManager>("outboxMongoTransactionManager"))
+
+    /**
+     * Provides a default Clock bean if none is configured.
+     *
+     * @return System default zone clock
+     */
+    @Bean
+    @ConditionalOnMissingBean
+    fun clock(): java.time.Clock = java.time.Clock.systemDefaultZone()
```
rolandbeisel: Could you use imports here, please

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxRecordEntity.kt
rolandbeisel: I removed the index

namastack-outbox-mongodb/src/main/kotlin/io/namastack/outbox/MongoOutboxRecordEntity.kt
rolandbeisel: I removed the index

rolandbeisel:
> @rolandbeisel Hi, I just finished refactoring the Mongo module to match JDBC parity. I've also built a demo project (Spring Boot 4 + Mongo 8) here: https://github.com/Stellarhold170NT/namastack-outbox-sample-mongodb. Let me know what you think

Thank you @Stellarhold170NT. I pushed two commits and polished some things. Please have a look :). Could you add an example project in namastack-outbox-examples?

rolandbeisel (approved):

rolandbeisel (dismissed):

rolandbeisel:
> Hi!
> 
> I have following questions:
> 
> 1. For relational databases, this library provides a way to customize table prefixes (in jdbc module). How about doing something like that for mongodb as well?
> 2. For relational dbs, there's a migration script for setting up schemas. MongoDB entities in here use annotations for indexes, but not every project allows creation of indexes based on that. How about adding in example a reference mongosh/js script for setting up collections and indexes?
> 
> Kind regards, Damian

@malczuuu @Stellarhold170NT I added support for customizing the collection prefix, along with an example setup script and the relevant documentation.

rolandbeisel (approved):

PR #237, opened 2026-03-13
GH-238 Fix handler ID instability with CGLIB proxies

yunhobb:
## Summary

- `BaseHandlerMethod.buildId()` uses `bean::class.java.name` directly, which returns the CGLIB proxy class name (e.g., `MyHandler$$SpringCGLIB$$0`) when AOP annotations like `@Transactional` are applied.
- This causes handler IDs stored in the database to be unstable — changing when AOP configuration changes (e.g., adding/removing `@Transactional`), which breaks retry policy matching via `retryPolicyRegistry.getByHandlerId()`.
- Fix: use the existing `ReflectionUtils.getTargetClass(bean).name` to resolve the actual target class behind AOP proxies.

## Changes

- **`BaseHandlerMethod.kt`**: Replace `bean::class.java.name` with `ReflectionUtils.getTargetClass(bean).name` in `buildId()`
- **`BaseHandlerMethodProxyTest.kt`**: Add tests verifying handler ID stability when beans are wrapped in CGLIB proxies

## Test plan

- [x] New tests pass: `BaseHandlerMethodProxyTest` — verifies proxy-wrapped beans produce the same handler ID as unwrapped beans
- [x] All existing tests pass: `./gradlew :namastack-outbox-core:test`
- [x] Code style checks pass: `./gradlew :namastack-outbox-core:ktlintCheck`


namastack-outbox-core/src/test/kotlin/io/namastack/outbox/handler/method/BaseHandlerMethodProxyTest.kt
```diff
@@ -0,0 +1,72 @@
...
+        assertThat(handlerFromProxy.id)
+            .describedAs(
+                "Handler ID from proxied bean should match target bean's ID. " +
+                    "Actual proxy ID: ${handlerFromProxy.id}, " +
+                    "Expected target ID: ${handlerFromTarget.id}",
+            ).isEqualTo(handlerFromTarget.id)
+    }
+
+    @Test
+    fun `handler ID should contain original class name not proxy class name`() {
```
Alek96: If I am not mistaken, `given` and `when` parts of the test are identical to the test above. If so, please merge them. 

And could you move the tests to the base test class - `TypedHandlerMethodTest`?

rolandbeisel: @Alek96 I moved the tests and did some polishing.

yunhobb:
@rolandbeisel Thank you for pointing out the breaking change. I missed the backward compatibility issue.
I have two ideas for the fix:

1. As you suggested, save both IDs (proxy class name + real class name) when registering handlers, so old records can still match
2. Add a fallback lookup that strips the CGLIB suffix (e.g. $$SpringCGLIB$$0) when the primary lookup fails

If it's okay, I'd like to try implementing one of these. Can I open a follow-up PR?

PR #240, opened 2026-03-14
Add legacy alias support for backward compatibility with CGLIB proxy handler IDs

yunhobb:
## Summary

Follow-up to #237. 

PR #237 fixed handler ID instability by using `ReflectionUtils.getTargetClass()` to resolve the actual target class behind CGLIB proxies. However, existing database records may still reference the old CGLIB proxy-based handler IDs (e.g.,`MyHandler$$SpringCGLIB$$0#handle(...)`).

This PR registers both stable and legacy handler IDs so that existing outbox records can still be resolved after upgrading.

### Context from #237
Handler IDs stored in the database were unstable because `buildId()` used the CGLIB proxy class name directly, which changes when AOP configuration changes.

## Changes

- **`OutboxHandlerRegistry`**: Add `registerAlias()` to register legacy ID in the ID-based lookup map only (does not duplicate into typed/generic handler lists)
- **`OutboxFallbackHandlerRegistry`**: Add `registerAlias()` to register legacy ID for fallback handler lookup
- **`OutboxRetryPolicyRegistry`**: Add `registerAlias()` to register legacy ID for retry policy lookup
- **`OutboxHandlerBeanPostProcessor`**: Add `buildLegacyId()` to detect CGLIB proxied beans and `registerLegacyAliases()` to register legacy IDs across all registries

### Note
Added `registerAlias()` to `OutboxRetryPolicyRegistry` as well. Although the maintainer discussion only mentioned handler registry and fallback registry, records with legacy handler IDs also need retry policy lookup to work correctly. Happy to remove if unnecessary.

## Test plan

- [x] `OutboxHandlerRegistryTest` — alias lookup, no duplication in typed lists, silent ignore on existing ID
- [x] `OutboxFallbackHandlerRegistryTest` — alias registration and lookup
- [x] `OutboxHandlerBeanPostProcessorTest` — legacy alias registration for CGLIB proxied beans, no alias for non-proxied beans, fallback alias registration
- [x] All existing tests pass: `./gradlew :namastack-outbox-core:build`

### DIFF
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
@@ -122,17 +122,19 @@ class OutboxHandlerRegistry {
      * were generated using CGLIB proxy class names. The alias allows those old records
      * to still find their handler after upgrading to stable (non-proxy) IDs.
      *
-     * Unlike [register], this method:
-     * - Only adds to the ID-based lookup map (not to typed/generic handler lists)
-     * - Silently ignores if the alias ID is already registered
+     * Unlike [register], this method only adds to the ID-based lookup map
+     * (not to typed/generic handler lists).
      *
      * @param aliasId The legacy handler ID to register as an alias
      * @param handlerMethod The handler method this alias should resolve to
+     * @throws IllegalStateException if the alias ID is already registered
      */
     internal fun registerAlias(
         aliasId: String,
         handlerMethod: OutboxHandlerMethod,
     ) {
-        handlersById.putIfAbsent(aliasId, handlerMethod)
+        check(handlersById.putIfAbsent(aliasId, handlerMethod) == null) {
+            "Duplicate alias ID detected: $aliasId"
+        }
     }
 }
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxFallbackHandlerRegistry.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxFallbackHandlerRegistry.kt
@@ -69,11 +69,14 @@ class OutboxFallbackHandlerRegistry {
      *
      * @param aliasId The legacy handler ID to register as an alias
      * @param fallbackHandlerMethod The fallback handler method this alias should resolve to
+     * @throws IllegalStateException if the alias ID is already registered
      */
     internal fun registerAlias(
         aliasId: String,
         fallbackHandlerMethod: OutboxFallbackHandlerMethod,
     ) {
-        fallbackHandlersByHandlerId.putIfAbsent(aliasId, fallbackHandlerMethod)
+        check(fallbackHandlersByHandlerId.putIfAbsent(aliasId, fallbackHandlerMethod) == null) {
+            "Duplicate fallback alias ID detected: $aliasId"
+        }
     }
 }
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/retry/OutboxRetryPolicyRegistry.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/retry/OutboxRetryPolicyRegistry.kt
@@ -66,12 +66,15 @@ class OutboxRetryPolicyRegistry(
      *
      * @param aliasId The legacy handler ID to register as an alias
      * @param policy The retry policy this alias should resolve to
+     * @throws IllegalStateException if the alias ID is already registered
      */
     fun registerAlias(
         aliasId: String,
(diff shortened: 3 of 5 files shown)


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
```diff
@@ -65,27 +68,63 @@ internal class OutboxHandlerBeanPostProcessor(
...
+            .forEach { policy -> retryPolicyRegistry.registerAlias(legacyId, policy) }
+    }
+
+    /**
+     * Builds a legacy handler ID using the bean's runtime class name (which may include
+     * CGLIB proxy suffixes like `$$SpringCGLIB$$0`).
+     *
+     * Returns null if the legacy ID would be identical to the stable ID (i.e., the bean is not proxied).
+     */
+    private fun buildLegacyId(handler: OutboxHandlerMethod): String? {
```
Alek96: I would move this method to `BaseHandlerMethod` so it is easy to see a difference between ID and legacy ID. Additionally, you could use it in tests, and not to recreate the `buildLegacyId` method there. 

Could you build the handler legacy Id the same way the handler Id is built?
```kotlin
protected fun buildLegacyId(): String {
    val className = bean::class.java.name
    val methodName = method.name
    val paramTypes = method.parameterTypes.joinToString(",") { it.name }

    return "$className#$methodName($paramTypes)"
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
```diff
@@ -114,4 +114,25 @@ class OutboxHandlerRegistry {
...
+     * to still find their handler after upgrading to stable (non-proxy) IDs.
+     *
+     * Unlike [register], this method:
+     * - Only adds to the ID-based lookup map (not to typed/generic handler lists)
+     * - Silently ignores if the alias ID is already registered
+     *
+     * @param aliasId The legacy handler ID to register as an alias
+     * @param handlerMethod The handler method this alias should resolve to
+     */
+    internal fun registerAlias(
```
Alek96, line 132: You forget to check for duplication

namastack-outbox-core/src/test/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessorTest.kt
```diff
@@ -283,4 +286,111 @@ class OutboxHandlerBeanPostProcessorTest {
...
+        }
+
+        private fun createCglibProxy(target: Any): Any {
+            val proxyFactory = ProxyFactory(target)
+            proxyFactory.isProxyTargetClass = true
+            proxyFactory.addAdvice(MethodInterceptor { it.proceed() })
+            return proxyFactory.proxy
+        }
+
+        private fun buildLegacyId(
```
Alek96: It would be better to not recreate the legacy Id builder method in a test

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
yunhobb: @Alek96 
I intentionally used `putIfAbsent` to silently keep the first entry, since alias registration is a secondary lookup path for existing DB records.

In practice, alias ID conflicts between different handlers are unlikely since legacy IDs are derived from each bean's unique runtime class + method signature. But I can add a safety check if you prefer.

Should I:
1. Keep current behavior (silently ignore all duplicates)
2. Throw only when the alias conflicts with a different handler
3. Throw on any duplicate, same as register()

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
yunhobb: Thanks for the suggestion! I moved buildLegacyId() to BaseHandlerMethod and updated tests to use handler.legacyId directly.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
Alek96: I would go with the option 3 - Throw on any duplicate, same as register. To make the the logic consists with normal handler registration

yunhobb: @Alek96 Updated all three registries and their tests!

Issue #51, opened 2025-10-27
SNS Integration


rolandbeisel:
Ready-made processors for common brokers: SNS

namastack-outbox-sns

PR #246, opened 2026-03-26
GH-51 SNS Integration Module

rolandbeisel:


namastack-outbox-sns/src/main/kotlin/io/namastack/outbox/sns/SnsOutboxHandler.kt
```diff
@@ -0,0 +1,141 @@
...
+    private val snsOperations: SnsOperations,
+    private val routing: SnsOutboxRouting,
+) : OutboxHandler {
+    private val logger = LoggerFactory.getLogger(SnsOutboxHandler::class.java)
+
+    override fun handle(
+        payload: Any,
+        metadata: OutboxRecordMetadata,
+    ) {
+        if (!routing.shouldExternalize(payload, metadata)) {
```
Alek96, line 77: I think, `shouldExternalize` from `OutboxRouting` may have a bug.

If we do not set the default route, then `shouldExternalize` would return true by default
```kotlin
fun shouldExternalize(
    payload: Any,
    metadata: OutboxRecordMetadata,
): Boolean = findRule(payload, metadata)?.filter(payload, metadata) ?: true
```

And `resolveTarget` then would throw `IllegalStateException`:
```kotlin
fun resolveTarget(
    payload: Any,
    metadata: OutboxRecordMetadata,
): String =
    findRule(payload, metadata)?.target(payload, metadata)
        ?: throw IllegalStateException("No routing rule found for payload type: ${payload::class.java.name}")

fun findRule(
    payload: Any,
    metadata: OutboxRecordMetadata,
): OutboxRoute? {
    for (rule in rules) {
        if (rule.matches(payload, metadata)) {
            return rule
        }
    }

    return defaultRule
}
```

Should `shouldExternalize` by default return false?

Alek96: I added a test that verifies my suspicion

Alek96: And I have added the fix - [Change shouldExternalize to return false by default](https://github.com/namastack/namastack-outbox/pull/246/commits/bab0ebfdef64eb0f4050c8af82705f89f734755c).
Feel free to revoke the fix if there is an error in my understanding

rolandbeisel: @Alek96 Thanks for the fix.

PR #266, opened 2026-04-18
added rowmappers to the repositories

dagframstad:
Some Oracle JDBC drivers have limited support for 'getObject(int, Class)' with column type: java.time.Instant.
Added RowMapper<T> implementations to avoid the following:

> DEBUG org.springframework.jdbc.support.JdbcUtils JDBC driver has limited support for 'getObject(int, Class)' with column type: java.time.Instant

Should also perform slightly better since exception throwing during database fetches is avoided in these cases.

### DIFF
--- a/namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxInstanceRepository.kt
+++ b/namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/JdbcOutboxInstanceRepository.kt
@@ -3,8 +3,10 @@ package io.namastack.outbox
 import io.namastack.outbox.instance.OutboxInstance
 import io.namastack.outbox.instance.OutboxInstanceRepository
 import io.namastack.outbox.instance.OutboxInstanceStatus
+import org.springframework.jdbc.core.RowMapper
 import org.springframework.jdbc.core.simple.JdbcClient
 import org.springframework.transaction.support.TransactionTemplate
+import java.sql.ResultSet
 import java.sql.Timestamp
 import java.time.Instant
 
@@ -27,6 +29,7 @@ internal open class JdbcOutboxInstanceRepository(
     private val tableNameResolver: JdbcTableNameResolver,
 ) : OutboxInstanceRepository {
     private val tableName = tableNameResolver.outboxInstance
+    private val rowMapper = JdbcOutboxInstanceEntityRowMapper()
 
     /**
      * Query to update an existing instance.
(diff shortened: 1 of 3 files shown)


rolandbeisel:
@dagframstad Thanks for your contribution. We'll review it.

rolandbeisel:
> Some Oracle JDBC drivers have limited support for 'getObject(int, Class)' with column type: java.time.Instant. Added RowMapper implementations to avoid the following:
> 
> > DEBUG org.springframework.jdbc.support.JdbcUtils JDBC driver has limited support for 'getObject(int, Class)' with column type: java.time.Instant
> 
> Should also perform slightly better since exception throwing during database fetches is avoided in these cases.

@dagframstad Does this debug log interfere with correct functionality, or is it only reporting a fallback path?

dagframstad:
> > Some Oracle JDBC drivers have limited support for 'getObject(int, Class)' with column type: java.time.Instant. Added RowMapper implementations to avoid the following:
> > > DEBUG org.springframework.jdbc.support.JdbcUtils JDBC driver has limited support for 'getObject(int, Class)' with column type: java.time.Instant
> > 
> > 
> > Should also perform slightly better since exception throwing during database fetches is avoided in these cases.
> 
> @dagframstad Does this debug log interfere with correct functionality, or is it only reporting a fallback path?

As far as I can tell everything works, but an exception is thrown every time the `Jdbc*Repository`-classes try to get an Instant from the `ResultSet`.
The Exception is thrown in `org.springframework.jdbc.support.JdbcUtils.getResultSetValue(ResultSet, int, Class<?>)`. 
```
			catch (SQLException ex) {
				if (logger.isDebugEnabled()) {
					logger.debug("JDBC driver has limited support for 'getObject(int, Class)' with column type: " +
							requiredType.getName(), ex);
				}
			}

			// Corresponding SQL types for JSR-310, left up to the caller to convert
			// them (for example, through a ConversionService).
			String typeName = requiredType.getSimpleName();
			return switch (typeName) {
				case "LocalDate" -> rs.getDate(index);
				case "LocalTime" -> rs.getTime(index);
				case "LocalDateTime" -> rs.getTimestamp(index);
				// Fall back to getObject without type specification, again
				// left up to the caller to convert the value if necessary.
				default -> getResultSetValue(rs, index);
			};
```

rolandbeisel:
Thanks @dagframstad! Could you tell me how to reproduce this exception?

dagframstad:
I'm using this:
`runtimeOnly("com.oracle.database.jdbc:ojdbc11:21.21.0.0")`
with this database:
`Oracle AI Database 26ai Free Release 23.26.1.0.0 - Develop, Learn, and Run for Free Version 23.26.1.0.0`
from a docker image `gvenzl/oracle-free:23-slim-faststart`.

My production database is an older version:
`Oracle Database 19c EE Extreme Perf Release 19.0.0.0.0 - Production Version 19.30.0.0.0`
but for now I'm only looking at this running locally.

Tried with `com.oracle.database.jdbc:ojdbc17:23.26.1.0.0` now. Same exception.

rolandbeisel (approved):

rolandbeisel:
@dagframstad Thank you for the contribution!

PR #268, opened 2026-04-22
GH-267 Add feature: time-based properties as java.time.Duration

### LINKED ISSUE #267: More intuitive configuration if delays, intervals and other time-based properties were specified as Duration
(Reading) the properties would be more intuitive if namastack-outbox used `java.time.Duration` and not `Long` for delays, intervals, etc.
At least my mind finds it easier to grasp _5m_ vs. _300000_, or _8h_ vs. _2880000_.
A breaking change, but a minor one.

dagframstad:
Should solve most of the things suggested in #267.
One breaking change I think: The `@Scheduled`annotation in `io.namastack.outbox.instance.OutboxInstanceRegistry`

### DIFF
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
@@ -63,14 +63,8 @@ class OutboxProcessingScheduler(
 
     private val log = LoggerFactory.getLogger(OutboxProcessingScheduler::class.java)
 
-    private val lifecycle =
-        SchedulerLifecycleStateMachine(
-            if (properties.processing.shutdownTimeoutSeconds != null) {
-                Duration.ofSeconds(properties.processing.shutdownTimeoutSeconds!!)
-            } else {
-                properties.processing.shutdownTimeout
-            },
-        )
+    private val lifecycle = SchedulerLifecycleStateMachine(properties.processing.shutdownTimeout)
+
     private var scheduledTask: ScheduledFuture<*>? = null
 
     /**
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
@@ -63,7 +63,7 @@ class OutboxProcessingScheduler(
 
     private val log = LoggerFactory.getLogger(OutboxProcessingScheduler::class.java)
 
-    private val lifecycle = SchedulerLifecycleStateMachine(properties.processing.shutdownTimeout)
+    private val lifecycle = SchedulerLifecycleStateMachine(properties.processing.effectiveShutdownTimeout)
 
     private var scheduledTask: ScheduledFuture<*>? = null
 
@@ -96,7 +96,7 @@ class OutboxProcessingScheduler(
      *
      * Automatically invoked during application shutdown. Cancels future scheduling and,
      * if a cycle is currently running, blocks until completion, timeout, or interruption
-     * (up to [OutboxProperties.Processing.shutdownTimeoutSeconds]).
+     * (up to [OutboxProperties.Processing.shutdownTimeout]).
      */
     override fun stop() {
         log.info("Initiating OutboxProcessingScheduler shutdown...")
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
@@ -12,6 +12,7 @@ import org.springframework.scheduling.TaskScheduler
 import org.springframework.scheduling.support.ScheduledMethodRunnable
 import java.lang.reflect.Method
 import java.time.Clock
+import java.time.Duration
 import java.util.concurrent.CountDownLatch
 import java.util.concurrent.ScheduledFuture
 import java.util.concurrent.TimeUnit
@@ -62,7 +63,14 @@ class OutboxProcessingScheduler(
 
     private val log = LoggerFactory.getLogger(OutboxProcessingScheduler::class.java)
 
-    private val lifecycle = SchedulerLifecycleStateMachine(properties.processing.shutdownTimeoutSeconds)
+    private val lifecycle =
+        SchedulerLifecycleStateMachine(
+            if (properties.processing.shutdownTimeoutSeconds != null) {
+                Duration.ofSeconds(properties.processing.shutdownTimeoutSeconds!!)
+            } else {
+                properties.processing.shutdownTimeout
+            },
+        )
     private var scheduledTask: ScheduledFuture<*>? = null
 
     /**
@@ -207,10 +215,10 @@ class OutboxProcessingScheduler(
      * It coordinates job registration, processing execution, and graceful shutdown,
      * including waiting for an in-flight cycle to complete with a bounded timeout.
      *
-     * @param shutdownTimeoutSeconds Maximum time to wait for an active cycle during shutdown
+     * @param shutdownTimeout Maximum time to wait for an active cycle during shutdown
      */
     class SchedulerLifecycleStateMachine(
-        private val shutdownTimeoutSeconds: Long,
+        private val shutdownTimeout: Duration,
     ) {
         /**
          * Lifecycle states for scheduler registration and processing execution.
@@ -331,14 +339,14 @@ class OutboxProcessingScheduler(
             }
 
         private fun awaitProcessingComplete() {
-            log.debug("Waiting for processing cycle to complete (timeout={}s)", shutdownTimeoutSeconds)
+            log.debug("Waiting for processing cycle to complete (timeout={}ms)", shutdownTimeout.toMillis())
 
             try {
-                val completed = processingComplete.await(shutdownTimeoutSeconds, TimeUnit.SECONDS)
+                val completed = processingComplete.await(shutdownTimeout.toMillis(), TimeUnit.MILLISECONDS)
                 if (completed) {
                     log.trace("Processing cycle completed while shutting down")
                 } else {
-                    log.warn("Shutdown timeout reached after {}s; forcing shutdown", shutdownTimeoutSeconds)
+                    log.warn("Shutdown timeout reached after {}ms; forcing shutdown", shutdownTimeout.toMillis())
                 }
             } catch (ex: InterruptedException) {
                 Thread.currentThread().interrupt()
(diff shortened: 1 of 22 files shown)


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instance/OutboxInstanceRegistry.kt
```diff
@@ -45,8 +45,18 @@ class OutboxInstanceRegistry(
 ) : SmartLifecycle {
     private val log = LoggerFactory.getLogger(OutboxInstanceRegistry::class.java)
 
-    private val staleInstanceTimeout = Duration.ofSeconds(properties.instance.staleInstanceTimeoutSeconds)
-    private val gracefulShutdownTimeout = Duration.ofSeconds(properties.instance.gracefulShutdownTimeoutSeconds)
+    private val staleInstanceTimeout =
+        if (properties.instance.staleInstanceTimeoutSeconds != null) {
```
rolandbeisel: I think this property is never null. We need:
```kotlin
private val staleInstanceTimeout =
        properties.instance.staleInstanceTimeout
            ?: Duration.ofSeconds(properties.instance.staleInstanceTimeoutSeconds)
```

This also applies to gracefulShutdownTimeout

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instance/OutboxInstanceRegistry.kt
```diff
@@ -146,7 +156,7 @@ class OutboxInstanceRegistry(
      * Combines update & pruning to reduce scheduling overhead.
      */
     @Scheduled(
```
rolandbeisel: To make this non-breaking, we should use TaskScheduler and not `@Scheduled`. Then we can calculate the fixedRateString like we do for `staleInstanceTimeout` and `gracefulShutdownTimeout`.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
```diff
@@ -62,7 +63,14 @@ class OutboxProcessingScheduler(
...
 
-    private val lifecycle = SchedulerLifecycleStateMachine(properties.processing.shutdownTimeoutSeconds)
+    private val lifecycle =
+        SchedulerLifecycleStateMachine(
+            if (properties.processing.shutdownTimeoutSeconds != null) {
+                Duration.ofSeconds(properties.processing.shutdownTimeoutSeconds!!)
+            } else {
+                properties.processing.shutdownTimeout
+            },
+        )
```
rolandbeisel: Elvis Operator looks cleaner in Kotlin imo

namastack-outbox-core/src/main/resources/META-INF/spring-configuration-metadata.json
```diff
@@ -271,39 +271,83 @@
       "description": "Concurrency limit for the SimpleAsyncTaskExecutor used for parallel processing of record keys when virtual threads are enabled. Default: -1 (no limit)."
     },
     {
-      "name": "namastack.outbox.processing.shutdown-timeout-seconds",
+      "name": "namastack.outbox.processing.shutdown-timeout-secods",
```
rolandbeisel: typo in the word "seconds"

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instance/OutboxInstanceRegistry.kt
dagframstad: This does not make any sense. 
`properties.instance.staleInstanceTimeout` is not nullable. `properties.instance.staleInstanceTimeoutSeconds` is now nullable, if `staleInstanceTimeoutSeconds` is set in the properties it should be used until the property is removed from the properties, at which point it will be null and `staleInstanceTimeout`will be used, or you decide to remove the deprecated property.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
dagframstad: do you mean like this?
```java
private val lifecycle =
    SchedulerLifecycleStateMachine(
        Duration.ofSeconds(
            properties.processing.shutdownTimeoutSeconds
                ?: properties.processing.shutdownTimeout.seconds,
        ),
    )
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instance/OutboxInstanceRegistry.kt
dagframstad: Will do.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
Alek96: What if we do the Elvis operator in the `OutboxProperties` class?
Then we could just use `properties.processing.shutdownTimeout` here:
```kotlin
private val lifecycle = SchedulerLifecycleStateMachine(properties.processing.shutdownTimeout)
```
`OutboxProperties`:
```kotlin
data class Processing(
    @Deprecated("Use shutdownTimeout instead")
    var shutdownTimeoutSeconds: Long? = null,
    private var shutdownTimeoutValue: Duration = Duration.ofSeconds(30),
) {
    var shutdownTimeout: Duration
        get() = shutdownTimeoutSeconds?.let(Duration::ofSeconds) ?: shutdownTimeoutValue
        set(value) {
            shutdownTimeoutValue = value
        }
}
```
I tested it, and Spring can accept both properties: `namastack.outbox.processing.shutdown-timeout` and `namastack.outbox.processing.shutdown-timeout-seconds`.

The same Idea can be applied to `properties.instance.staleInstanceTimeout` and `properties.instance.gracefulShutdownTimeout`, simplifying `OutboxInstanceRegistry` as well.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProperties.kt
```diff
@@ -93,7 +97,10 @@ data class OutboxProperties(
         var executorCorePoolSize: Int = 4,
         var executorMaxPoolSize: Int = 8,
         var executorConcurrencyLimit: Int = -1,
-        var shutdownTimeoutSeconds: Long = 30,
+        @Deprecated("Use shutdownTimeout instead")
+        var shutdownTimeoutSeconds: Long? = null,
+        @DurationUnit(ChronoUnit.SECONDS)
```
Alek96: I’m not sure adding the duration unit here is a good idea. It could be wise to use a single unit throughout. By doing so, we would establish a clear standard.

Same comment for `heartbeatInterval`, `staleInstanceTimeout`, and `gracefulShutdownTimeout`.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instance/OutboxInstanceRegistry.kt
Alek96: Or we can use SpEL:
```kotlin
@Scheduled(
    fixedRateString = 
        $$"#{${namastack.outbox.instance.heartbeat-interval-seconds:null} != null ? ${namastack.outbox.instance.heartbeat-interval-seconds:null} * 1000 : ${namastack.outbox.instance.heartbeat-interval:5000}}",
)
fun performHeartbeatAndCleanup() {
```

[I did not test it]

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProperties.kt
dagframstad: Agree.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
dagframstad: @Alek96 great idea, will do,

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instance/OutboxInstanceRegistry.kt
```diff
@@ -141,15 +144,17 @@ class OutboxInstanceRegistry(
         }
     }
 
+    @PostConstruct
+    fun scheduleHeartbeat() {
+        val rate = properties.instance.heartbeatInterval
+        taskScheduler.scheduleAtFixedRate({ performHeartbeatAndCleanup() }, rate)
```
Alek96: To make the scheduled job observable, you have to wrap it with `ScheduledMethodRunnable` just like it was done in `OutboxProcessingScheduler`

dagframstad: But was it observable before?

Alek96: Yes, anything annotated with @Scheduled will be observable by default.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/config/OutboxCoreInfrastructureAutoConfiguration.kt
```diff
@@ -68,7 +71,16 @@ class OutboxCoreInfrastructureAutoConfiguration {
         instanceRepository: OutboxInstanceRepository,
         properties: OutboxProperties,
         clock: Clock,
-    ): OutboxInstanceRegistry = OutboxInstanceRegistry(instanceRepository, properties, clock)
+        taskScheduler: TaskScheduler,
```
Alek96: You are using a default `TaskScheduler` here instead of the `outboxHeartbeatScheduler`. 

As the desired scheduler cannot be injected normally, you should do it like this: 
```kotlin
val taskScheduler = beanFactory.getBean(OutboxProcessingScheduler.SCHEDULER_NAME) as TaskScheduler
```

rolandbeisel:
@dagframstad Thanks for your work. I pushed an additional commit... there was a double scheduling happening the OutboxInstanceRegistry, which I fixed. I would like to test the migration and backwards-compatibility of these changes manually with the help of some example projects before merging this branch.

rolandbeisel (dismissed):

rolandbeisel:
@dagframstad I’ve pushed a new commit.

This update refactors the time-based configuration properties — `shutdownTimeout`, `heartbeatInterval`, `staleInstanceTimeout`, and `gracefulShutdownTimeout` - so that they are now defined as constructor parameters in the configuration classes. This change ensures proper Spring Boot autocompletion and metadata generation.

Previously, autocompletion would suggest properties like `shutdownTimeoutValue`, which was not ideal.

I’ve also updated and completed the documentation and KDoc.

rolandbeisel (approved):

PR #273, opened 2026-04-24
Add handler-level supports hook and filter handlers before scheduling outbox records

### LINKED ISSUE #274: Add scheduling-time handler support checks for generic outbox handlers
## Problem
 
Messaging integrations (Kafka/SNS/Rabbit) are registered as generic `OutboxHandler`s. During scheduling, the outbox currently creates one record per discovered handler, even when the handler will later skip the payload based on routing filter/configuration.
 
This causes avoidable outbox records and unnecessary processing overhead.
 
## Goal
 
Introduce a scheduling-time support check so handlers can opt out *before* records are persisted.
 
## Proposed Solution
 
1. Extend `OutboxHandler` with:
- `supports(payload: Any, metadata: OutboxRecordMetadata): Boolean`
- default implementation returns `true` to preserve backward compatibility.
2. During handler collection in `OutboxService`, evaluate support before persisting a record for a handler.
3. Implement `supports(...)` in messaging handlers (Kafka, SNS, Rabbit) and delegate to routing/filter logic.
4. Keep runtime safety guard in `handle(...)` for defensive behavior.
 
## Backward Compatibility
 
- Existing custom handlers remain compatible because `supports(...)` defaults to `true`.
- Existing typed handlers are unaffected.
 
## Acceptance Criteria
 
- [ ] Outbox does not persist records for generic handlers when `supports(...)` is `false`.
- [ ] Kafka/SNS/Rabbit handlers expose `supports(...)` and align with routing `filter` outcomes.
- [ ] Existing handlers compile and behave as before when they do not override `supports(...)`.
- [ ] Unit tests cover positive/negative support cases.
 
## Notes
 
This improves performance and reduces noise in outbox tables, especially for applications with multiple enabled messaging modules.

rolandbeisel:
### Motivation

- Allow handlers to opt-out of scheduling so outbox records are not created when a handler does not want to process a payload. 
- Use routing/filter logic consistently at both scheduling and handling time to avoid unnecessary record creation and duplicate filtering work. 
- Preserve backward compatibility by defaulting new hooks to allow scheduling.

### Description

- Adds `supports(payload, metadata)` to the `OutboxHandler` interface with a default `true` implementation. 
- Adds `supportsScheduling(payload, metadata)` to `OutboxHandlerMethod` with a default `true`, and implements it in `GenericHandlerMethod` to call `OutboxHandler.supports` on the bean when available. 
- Updates `OutboxService.collectHandlers` to construct `OutboxRecordMetadata` (including `createdAt` from the service `clock`) and filter discovered handlers by `supportsScheduling` before creating records. 
- Updates `KafkaOutboxHandler`, `RabbitOutboxHandler`, and `SnsOutboxHandler` to implement `supports` by delegating to their routing `shouldExternalize` and to use `supports` inside `handle` for early return. 
- Updates/extends unit tests and test helpers to assert the new behavior and to mock `supportsScheduling` where needed.

### Testing

- Ran unit tests with `./gradlew test` covering `OutboxServiceTest`, `KafkaOutboxHandlerTest`, `RabbitOutboxHandlerTest`, and `SnsOutboxHandlerTest`, including the new cases `should not create record for generic handler when supports returns false` and `supports returns false when filter excludes payload`. 
- All modified and newly added unit tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69eb54e939608326811faa596f306502)


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/handler/GenericHandlerMethod.kt
```diff
@@ -18,6 +19,11 @@ class GenericHandlerMethod(
     bean: Any,
     method: Method,
 ) : OutboxHandlerMethod(bean, method) {
+    override fun supportsScheduling(
+        payload: Any,
+        metadata: OutboxRecordMetadata,
+    ): Boolean = (bean as? OutboxHandler)?.supports(payload, metadata) ?: true
```
Alek96: I would move the `supports` check into `getGenericHandlers()`. 

Right now we are returning all genetic handlers and type handlers only to cast them to `OutboxHandler` a second later, it is making a mess.

I will test it, and see if it will make the code cleaner

Alek96:
I have published a commit that moves the filtering of GenericHandlerMethod into OutboxHandlerRegistry.
Feel free to change it or revoke it

rolandbeisel:
> I have published a commit that moves the filtering of GenericHandlerMethod into OutboxHandlerRegistry. Feel free to change it or revoke it

Thanks @Alek96 Looks good to me. Would you approve the PR?

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/handler/GenericHandlerMethod.kt
rolandbeisel: Looks good!

Alek96 (approved):

PR #276, opened 2026-04-25
Add partition lease ownership and draining handoff

arun0009:
## Summary
- add lease-based partition ownership with renewal/expiry and claimable-processable semantics
- introduce draining-aware rebalance using in-flight tracking to release partitions only after local work completes
- persist new partition fields across JDBC/JPA/Mongo, update SQL schemas, and add additive Flyway `V3` migrations in examples
- document `namastack.outbox.partition.lease-duration-seconds` and lease/draining behavior

## Why
In autoscaling/container environments (for example AWS Fargate), frequent instance churn can leave partition ownership stale or handoff abrupt. Lease-based ownership provides bounded failover, and draining improves handoff safety by waiting for local in-flight work to finish before release.

## Breaking change
- database migration is required before upgrade (`outbox_partition.lease_expires_at`, `outbox_partition.draining`); treat this as a major-version change

## Test plan
- [x] existing core unit/integration tests updated for claimable/draining semantics
- [ ] run full CI matrix for JDBC/JPA/Mongo modules
- [ ] validate Flyway upgrade path from `V1`+`V2` to `V3` in example apps

### DIFF
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxProcessingScheduler.kt
@@ -3,6 +3,8 @@ package io.namastack.outbox
 import io.micrometer.observation.ObservationRegistry
 import io.namastack.outbox.OutboxRecordStatus.NEW
 import io.namastack.outbox.partition.PartitionCoordinator
+import io.namastack.outbox.partition.PartitionDrainTracker
+import io.namastack.outbox.partition.PartitionHasher
 import io.namastack.outbox.processor.OutboxRecordProcessor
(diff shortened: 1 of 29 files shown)


rolandbeisel:
@arun0009 Thanks for the proposal and the detailed write-up.

At the moment, this change set (lease-based ownership, draining handoff, and required DB migration) feels too large to introduce directly via PR without prior alignment.

Given the scope and operational impact, this should first be discussed in an Issue/Discussion so we can agree on design, rollout strategy, migration plan, and versioning expectations before implementation.

Our Contribution Guidelines explicitly ask contributors to open an issue to discuss significant changes before starting work and to check existing issues/discussions first.

Please open a discussion first, and then we can decide together whether and how this should move forward as a PR. Thanks!

PR #289, opened 2026-05-08
GH-288 add JPA schema validation tests for all supported databases and fix DDL compatibility issues

### LINKED ISSUE #288: The JPA entity `io.namastack.outbox.OutboxRecordEntity` declares `payload` and `context` as plain `String` without `@Lob`
## Summary

The JPA entity `io.namastack.outbox.OutboxRecordEntity` (in `namastack-outbox-jpa` / `namastack-outbox-starter-jpa`) declares `payload` and `context` as plain `String` properties, with neither `@Lob` nor an explicit length:

```kotlin
@Column(name = "payload", nullable = false)
val payload: String,

@Column(name = "context", nullable = true)
val context: String?,
```

However, the Oracle DDL shipped by the project ([`namastack-outbox-jdbc/src/main/resources/schema/oracle/outbox-tables.sql`](https://github.com/namastack/namastack-outbox/blob/main/namastack-outbox-jdbc/src/main/resources/schema/oracle/outbox-tables.sql)) creates both columns as `NCLOB`:

```sql
payload  NCLOB NOT NULL,
context  NCLOB,
```

Without `@Lob`, Hibernate maps a plain `String` to `VARCHAR(255)` and refuses to validate against an `NCLOB` column. Even with `@Lob`, Hibernate maps `String` to `CLOB` by default, not `NCLOB` — so `@Nationalized` is also required for the entity to align with the shipped Oracle schema. As a result, applications that use the official DDL together with `hibernate.ddl-auto=validate` (the recommended production setting) cannot start.

## Environment

- Spring Boot 4.0.4 / Spring Kafka 4.0.x
- Hibernate ORM 7.x (Jakarta Persistence 3.2)
- Oracle 19c, schema created from the project's `schema/oracle/outbox-tables.sql`
- `spring.jpa.hibernate.ddl-auto: validate`

## Steps to reproduce

1. Add `namastack-outbox-starter-jpa:1.4.1` to a Spring Boot 4 application.
2. Apply the official Oracle DDL from `schema/oracle/outbox-tables.sql` (e.g. via Liquibase) to create the `outbox_record` table.
3. Configure JPA with `spring.jpa.hibernate.ddl-auto=validate`.
4. Start the application.

## Actual behavior

Startup fails with:

```
org.hibernate.tool.schema.spi.SchemaManagementException:
  Schema validation: wrong column type encountered in column [context]
  in table [outbox_record]; found [nclob (Types#NCLOB)],
  but expecting [varchar2(255 char)]
```

The same error then occurs for `payload` once `context` is fixed.

## Expected behavior

The shipped DDL and the JPA entity should agree on the column type, so an application using both can start with `ddl-auto=validate`.

## Proposed fix

Add `@Lob` and `@Nationalized` to both fields in `OutboxRecordEntity`, so Hibernate maps them to `NCLOB` (Oracle) / `NTEXT`-equivalents on other vendors, matching the shipped Oracle schema:

```kotlin
@Lob
@Nationalized
@Column(name = "payload", nullable = false)
val payload: String,

@Lob
@Nationalized
@Column(name = "context", nullable = true)
val context: String?,
```

Notes:

- `@Nationalized` is a Hibernate annotation (`org.hibernate.annotations.Nationalized`). The same effect can be achieved globally via `hibernate.use_nationalized_character_data=true`, but a per‑field annotation keeps the entity self‑describing and doesn't impose a global Hibernate setting on the host application.
- `@Basic(fetch = FetchType.LAZY)` could optionally be added if you'd like to keep large payloads off the default fetch path.
- Cross‑check that the other shipped schemas (`schema/postgresql/`, `schema/mysql/`, `schema/h2/`, …) also use national/Unicode column types so the entity and DDL stay consistent across all supported databases. If some vendor scripts use plain `TEXT`/`CLOB`, they should be aligned with Oracle's `NCLOB` choice (or vice‑versa).

## Workaround (for users hitting this today)

Place a `META-INF/orm.xml` on the classpath to augment the upstream entity mapping without forking the library. JPA XML has no element for `@Nationalized`, so we pin the column definition explicitly to `NCLOB`:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<entity-mappings xmlns="https://jakarta.ee/xml/ns/persistence/orm"
                 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
                 xsi:schemaLocation="https://jakarta.ee/xml/ns/persistence/orm
                                     https://jakarta.ee/xml/ns/persistence/orm/orm_3_0.xsd"
                 version="3.0">
    <entity class="io.namastack.outbox.OutboxRecordEntity" access="FIELD">
        <attributes>
            <basic name="payload">
                <column name="payload" nullable="false" column-definition="NCLOB"/>
            </basic>
            <basic name="context">
                <column name="context" nullable="true" column-definition="NCLOB"/>
            </basic>
        </attributes>
    </entity>
</entity-mappings>
```

rolandbeisel:
# JPA Schema Validation Fixes

This PR fixes JPA schema validation failures for `OutboxRecordEntity.payload` and
`OutboxRecordEntity.context`.

Previously, these fields were mapped as plain `String` values, so Hibernate expected
`VARCHAR(255)`. This did not match the shipped DDL schemas for large payload/context
columns, most visibly on Oracle where the official schema uses `NCLOB`. Applications
using the shipped DDL with `spring.jpa.hibernate.ddl-auto=validate` could fail during
startup.

The fields now use Hibernate's dialect-aware large Unicode text mapping.

## Impact

| Setup                                            | Impact                               |
|--------------------------------------------------|--------------------------------------|
| JDBC module                                      | No runtime migration required        |
| JPA with `ddl-auto=none`                         | No startup validation impact         |
| JPA with `ddl-auto=validate` and new schema      | No action required                   |
| JPA with `ddl-auto=validate` and existing schema | Migration may be required, see below |

## Database Notes

| Database   | Change                                                              |
|------------|---------------------------------------------------------------------|
| Oracle     | No DDL change. Existing `NCLOB` columns now validate correctly.     |
| PostgreSQL | No DDL change. Existing `TEXT` columns now validate correctly.      |
| SQL Server | `context` should be `NVARCHAR(MAX)` instead of `VARCHAR(MAX)`.      |
| MySQL      | `payload` and `context` should be `LONGTEXT CHARACTER SET utf8mb4`. |
| MariaDB    | `payload` and `context` should be `LONGTEXT CHARACTER SET utf8mb4`. |
| H2         | `payload` and `context` should be `CLOB`.                           |

### SQL Server

```sql
ALTER TABLE outbox_record
    ALTER COLUMN context NVARCHAR(MAX) NULL;
```

### MySQL

```sql
ALTER TABLE outbox_record
    MODIFY COLUMN payload LONGTEXT CHARACTER
SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
    MODIFY COLUMN context LONGTEXT CHARACTER
SET utf8mb4 COLLATE utf8mb4_unicode_ci NULL;
```

### MariaDB

```sql
ALTER TABLE outbox_record
    MODIFY COLUMN payload LONGTEXT CHARACTER
SET utf8mb4 COLLATE utf8mb4_unicode_ci NOT NULL,
    MODIFY COLUMN context LONGTEXT CHARACTER
SET utf8mb4 COLLATE utf8mb4_unicode_ci NULL;
```

### H2

H2 is typically used for tests and local development. Recreate the schema from the updated
DDL, or alter the columns manually:

```sql
ALTER TABLE outbox_record
    ALTER COLUMN payload CLOB;
ALTER TABLE outbox_record
    ALTER COLUMN context CLOB;
```


namastack-outbox-integration-tests/build.gradle.kts
```diff
@@ -19,6 +20,19 @@ dependencies {
     testImplementation(libs.mockk)
     testImplementation(libs.h2)
     testImplementation(libs.junit.jupiter)
+
+    testImplementation(libs.testcontainers.junit.jupiter)
```
Alek96, line 24: Currently, we are using the `1.21.4` version of `testcontainers`, but the latest version is [2.0.5](https://mvnrepository.com/artifact/org.testcontainers/testcontainers/2.0.5). I can add a commit that bumps the versions and relevant libraries

Alek96 (dismissed):

namastack-outbox-integration-tests/build.gradle.kts
rolandbeisel: Fixed! Thanks for mentioning.

Alek96 (approved):

PR #297, opened 2026-05-20
Add a test to each example to verify if the examples are working

### LINKED ISSUE #298: Add Smoke Tests

Alek96:


.github/workflows/example-smoke-tests.yml
```diff
@@ -0,0 +1,101 @@
...
+    branches:
+      - 'main'
+      - 'release/**'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  publish-snapshot:
```
Alek96, line 14: I was considering a slightly different approach. To create a GitHub workflow that would run only the smoke tests, that could be triggered by hand and by other workflow. Then the publish workflow could trigger a smoke test workflow. 

Your solution is also fine. I just wanted to mention an alternative solution

rolandbeisel: Not sure, if I understand you correctly. My idea is to run this workflow on every push to main and release branches, and optionally allow it to be triggered manually at any time.

The workflow only runs the smoke tests, but it first needs to build the library and publish it to the local Maven repository. This ensures that each example uses the current and correct dependencies for the Namastack modules.

Alek96: Now I see that I did not fully understand your idea. I did not see that it was publishing to the local Maven repository. 

I agree that we can run it on every pull request, not on every release.

PR #359, opened 2026-05-22
Include cause in log when failed to bootstrap partitions

tomwetjens:
We have a MongoDB database and get an error in our logs:
```
Failed to bootstrap partitions for instance de1808e0-d789-430b-9d80-2853cf6e2c75: Could not commit Mongo transaction for session
```

I would like the cause exception logged so it is easier to debug this.


rolandbeisel (approved):

rolandbeisel:
@tomwetjens Thanks for the contribution. Do you already know what the problem is?

tomwetjens:
@rolandbeisel I suspect it was a transient issue in the MongoDB cluster and nothing related to the Namastack Outbox library. I have stopped and started our application multiple times and the error has not come back. 
But adding the cause is good just in case, when future errors might occur.

PR #366, opened 2026-05-31
Add change to decouple Records and RecordHandlers from FQCN

### LINKED ISSUE #361: Feature Request: Logical name support in `@OutboxEvent` to decouple `record_type` from Java FQCN
# Feature Request: Logical name support in `@OutboxEvent` to decouple `record_type` from Java FQCN

## Problem

`record_type` in `outbox_record` is written as `payload.getClass().getName()` (hardcoded
in `OutboxRecordEntityMapper.map(OutboxRecord)`).  On the read path, `resolveClass()`
calls `ClassLoader.loadClass(recordType)`, which only works for fully-qualified class
names.

This creates a hard coupling between the Java package structure and the persisted outbox
rows: **renaming or moving an event class breaks deserialization of any `PENDING` or
`FAILED` rows written before the deploy.**  The poller throws `ClassNotFoundException`
and those rows enter permanent retry failure.

The only workarounds today are:

1. Drain the outbox for the affected type before deploying (requires a maintenance window
   or careful coordination).
2. Ship a Flyway `UPDATE outbox_record SET record_type = '<new FQCN>' WHERE record_type =
   '<old FQCN>'` in the same deploy.
3. Keep a deprecated subclass shim at the old FQCN for one release cycle.

All three add operational or code overhead for what should be a routine refactor.

---

## Proposed solution

Add an optional `name` attribute to `@OutboxEvent`:

```java
@OutboxEvent(key = "#root.id.toString()", name = "SiteEvent")
public class SiteEvent { … }
```

When `name` is provided, store it in `record_type` instead of the FQCN.  The name is a
stable, human-readable contract that survives package reorganisations.

---

### 1. `@OutboxEvent` annotation — add `name` attribute

```kotlin
annotation class OutboxEvent(
    val key: String,
    val name: String = "",          // empty = fall back to FQCN (backward compat)
    val context: Array<OutboxContextEntry> = []
)
```

### 2. Write path — `OutboxRecordEntityMapper.map(OutboxRecord<?>)`

Replace:

```kotlin
val recordType = record.payload!!.javaClass.name
```

With:

```kotlin
val ann = record.payload!!.javaClass.getAnnotation(OutboxEvent::class.java)
val recordType = if (ann != null && ann.name.isNotEmpty()) ann.name
                 else record.payload!!.javaClass.name
```

### 3. Name → class registry

A new `OutboxEventTypeRegistry` (a `Map<String, KClass<*>>`) built at startup.  The
natural population point is `OutboxHandlerBeanPostProcessor.postProcessAfterInitialization`,
which already iterates every Spring bean and introspects handler method parameter types.
For each discovered parameter type that carries `@OutboxEvent(name = "...")`, register
`name → class`:

```kotlin
fun register(payloadType: KClass<*>) {
    val ann = payloadType.findAnnotation<OutboxEvent>() ?: return
    if (ann.name.isEmpty()) return
    val existing = registry.putIfAbsent(ann.name, payloadType)
    check(existing == null || existing == payloadType) {
        "Duplicate @OutboxEvent name '${ann.name}': " +
        "registered for both ${existing!!.qualifiedName} and ${payloadType.qualifiedName}"
    }
}
```

Duplicate names fail fast at startup (`BeanCreationException`), preventing silent
misrouting.

### 4. Read path — `OutboxRecordEntityMapper.map(OutboxRecordEntity)`

Extend `resolveClass` to consult the registry first, falling back to `ClassLoader.loadClass`
for rows written before this feature was adopted (backward compatibility with existing
`PENDING`/`FAILED` rows that still store a FQCN):

```kotlin
private fun resolveClass(recordType: String): Class<*> =
    registry.find(recordType)?.java          // logical name lookup
        ?: Thread.currentThread()
               .contextClassLoader
               .loadClass(recordType)        // FQCN fallback — old rows still work
```

---

## Migration story

| Scenario | What happens |
|---|---|
| Existing rows with FQCN, library upgraded | `resolveClass` falls back to `loadClass(FQCN)` — no breakage |
| New rows written after upgrade, `name` set | `record_type` stores logical name; rows survive any future package move |
| New rows written after upgrade, no `name` | `record_type` stores FQCN; existing behavior unchanged |
| Class moved, rows with old FQCN still pending | Still broken — but this was already broken before this feature. The feature prevents *new* rows from being affected |

No migration SQL is required on adoption; the fallback handles old rows automatically.

---

## Why `OutboxHandlerBeanPostProcessor` is the right registry build point

The post-processor already:
- Scans every Spring bean in the context.
- Introspects `@OutboxHandler` method parameter types (to determine what event type each
  handler handles).
- Runs before the outbox poller starts polling.

Registering `@OutboxEvent(name)` as a side effect of handler scanning means:
- No separate classpath scan is needed.
- Only event types that actually have a handler registered are in the registry — exactly
  the set that can ever appear in `record_type`.
- The duplicate-name check runs at the same time as other startup validation.

---

## Example — before and after a package move

**Before (today):**

```
record_type = "com.example.events.SiteEvent"
```

Moving `SiteEvent` to `com.example.events.site.SiteEvent` → rows with old FQCN fail.

**After (with this feature):**

```java
@OutboxEvent(key = "#root.id.toString()", name = "SiteEvent")
public class SiteEvent { … }
```

```
record_type = "SiteEvent"
```

Moving the class to any package → `record_type` unchanged → no rows affected.

isarang:
Fixes #361


rolandbeisel:
Thanks for the contribution. As discussed in the issue, stable IDs may be worth considering as part of a broader persisted-contract design for a possible v2. We are not planning to introduce this change at this time.
The current compatibility migration approach remains the recommended solution. 

Please discuss significant implementation work in an issue first, as described in our contribution guidelines.

PR #368, opened 2026-05-31
feat: Add support for per-payload serializer: @OutboxEvent(serializer = ...)

### LINKED ISSUE #367: Feature Request: Per-payload serializer SPI: @OutboxEvent(serializer = ...)
### Problem

`OutboxPayloadSerializer` is a global swap — one serializer for all events. Teams using mixed
formats have no way to configure serialization per payload type without replacing Jackson
entirely. The existing docs demonstrate this pattern: a custom serializer bean silently replaces
the Jackson default for every event.

---

### Proposed solution

Add a `serializer` attribute to `@OutboxEvent` so format can be declared alongside the event
itself:

```kotlin
// Uses global Jackson default (unchanged)
@OutboxEvent(key = "#this.orderId")
data class OrderCreatedEvent(val orderId: String, val customerId: String)

// Uses a custom serializer — declared at the event, not globally
@OutboxEvent(key = "#this.orderId", serializer = ProtobufOutboxPayloadSerializer::class)
data class OrderShippedEvent(val orderId: String, val destination: String)

Jackson remains the global default for all unannotated events. No schema changes required —
the read path derives the correct serializer from record_type.

---
Design

- OutboxSerializerRegistrar — scans @OutboxEvent classes at startup (respects
namastack.outbox.event-scan-packages), validates that every declared serializer bean exists,
and fails fast with a clear message if not
- OutboxPayloadSerializerRegistry — routes forType(clazz) to the configured serializer;
falls back to Jackson for unconfigured types
- Context maps always use the default serializer (they carry no type annotation)

---
Required config when mixing serializers

Because any OutboxPayloadSerializer bean suppresses Jackson's auto-configuration, both must
be declared explicitly:

@Configuration
class SerializerConfig {
    @Bean @Primary
    fun jacksonSerializer(): OutboxPayloadSerializer = JacksonOutboxPayloadSerializer(...)

    @Bean  // registered by concrete type — picked up by OutboxSerializerRegistrar
    fun protobufSerializer(): ProtobufOutboxPayloadSerializer = ProtobufOutboxPayloadSerializer()
}

isarang:
Fixes #367


rolandbeisel:
Thanks for your interest in Namastack Outbox. I’m closing this PR because the proposed serializer SPI is not needed. OutboxPayloadSerializer is intended for database persistence; outbound payloads can already be transformed with mapPayload or a custom handler.

For significant changes, please open an issue first as described in our contribution guidelines.

PR #379, opened 2026-06-04
Allow package-private annotated methods

### LINKED ISSUE #381: Use MethodIntrospector to support package-private annotated handler methods
## Context

The handler discovery code previously relied on `Class.methods`, which only exposes public methods. That means Java package-private methods annotated with outbox annotations were not discovered, even though such methods are valid user-declared handler methods.

## Implementation

Update `ReflectionUtils` to use Spring's `MethodIntrospector` for method discovery. This allows the scanner to inspect user-declared methods on the target class, including Java package-private methods, and then apply the existing annotation lookup/accessibility handling.

This affects discovery for:
- `@OutboxHandler`
- `@OutboxFallbackHandler`
- `@OutboxRetryable` when present on a discovered handler method

## Expected Behavior

Java handlers should be able to declare package-private annotated methods, for example:

```java
public class CustomerHandler {

    @OutboxHandler
    void handle(CustomerEvent payload, OutboxRecordMetadata metadata) {
        // handle event
    }
}
```

The method should be discovered, registered, and usable like a public annotated handler method.

## Verification

Add or keep tests covering package-private Java annotated methods for:
- Reflection discovery
- Handler registration
- Fallback registration
- Retry policy registration

damianmalczewski:
Annotations such as `@KafkaListener`, `@RabbitListener`, etc. from Spring itself are applied on a method even if it's `package-private`. Right now, this library requires all methods annotated with `@OutboxHandler` to be `public`, due to using Java's native `Class.getMethods()` instead of Spring's `MethodIntrospector`. I propose to keep it the same way, to avoid confusion and improve Java interop of this library.

I know that as in `CONTRIBUTING.md`, the project is 100% Kotlin, but I had to add Java classes in test sources to properly test package-private methods.

### DIFF
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/internal/ReflectionUtils.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/internal/ReflectionUtils.kt
@@ -1,6 +1,7 @@
 package io.namastack.outbox.handler.method.internal
 
 import org.springframework.aop.support.AopUtils
+import org.springframework.core.MethodIntrospector
 import org.springframework.core.annotation.AnnotatedElementUtils
 import java.lang.reflect.Method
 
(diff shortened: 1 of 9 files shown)


rolandbeisel:
@damianmalczewski Thank you for your contribution. Looks good to me.

PR #408, opened 2026-07-21
GH-404 Support fully configurable table names in JDBC module

### LINKED ISSUE #404: Support fully configurable table names in JDBC module
### Problem

The JDBC module resolves table names as `${tablePrefix}` + a hardcoded lowercase base name (`outbox_record`, `outbox_instance`, `outbox_partition`) in `JdbcTableNameResolver`. The resolver is a `final` class and the base names are not configurable, so the prefix introduced in #130 / #152 is the only naming lever available.

### Motivation

In corporate environments, database object naming is often mandated by organization-wide standards (DBA/compliance-enforced catalogs), commonly requiring a fixed pattern such as all-uppercase identifiers (`ACME_OUTBOX_RECORD`) inherited from Oracle conventions. With MySQL on Linux (`lower_case_table_names=0`), identifiers are case-sensitive, so a prefix alone cannot satisfy such standards — the best achievable today is a mixed-case name like `ACME_outbox_record`, which fails naming-convention reviews. Since schema initialization is already expected to be manual (Flyway/Liquibase) when custom naming is used, full name control on the query side is the missing piece for standardized deployments.

### Proposal

Make the base table names configurable with the current values as defaults, e.g.:

```yaml
namastack:
  outbox:
    jdbc:
      table-prefix: "ACME_"
      table-names:
        record: "OUTBOX_RECORD"
        instance: "OUTBOX_INSTANCE"
        partition: "OUTBOX_PARTITION"
```

Alternatively (or additionally), opening `JdbcTableNameResolver` for extension would let users override resolution entirely, since the auto-configuration already registers it behind `@ConditionalOnMissingBean`.

Fully backward compatible: defaults preserve today's names. I'd expect the same guardrail as custom prefixes — automatic schema initialization unsupported with custom names.

Alek96:
This PR presents two commits for the same change — the first uses Kotlin val properties on the JdbcTableNameResolver interface, the second uses explicit getX() methods; @rolandbeisel please pick the option you prefer and I'll drop the other commit.

### DIFF
--- a/namastack-outbox-docs/docs/reference/configuration.md
+++ b/namastack-outbox-docs/docs/reference/configuration.md
@@ -53,8 +53,12 @@ namastack:
       rebalance-interval: 10s                  # How often partitions are recalculated (default: 10s)
 
     jdbc:
-      table-prefix: ""                         # Prefix for table names (default: empty)
       schema-name: null                        # Database schema name (default: null, uses default schema)
+      table-prefix: ""                         # Prefix for table names (default: empty)
+      table-names:
+        record: "outbox_record"                # Base name for the records table (default: outbox_record)
+        instance: "outbox_instance"            # Base name for the instances table (default: outbox_instance)
+        partition: "outbox_partition"          # Base name for the partitions table (default: outbox_partition)
       schema-initialization:
         enabled: true                          # Auto-create tables on startup (default: true)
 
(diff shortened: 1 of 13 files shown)


rolandbeisel:
> This PR presents two commits for the same change — the first uses Kotlin val properties on the JdbcTableNameResolver interface, the second uses explicit getX() methods; @rolandbeisel please pick the option you prefer and I'll drop the other commit.

@Alek96 Thank you! I would prefer the first one.

rolandbeisel (dismissed):
Looks good! Thank you!

rolandbeisel (approved):

PR #428, opened 2026-08-19
GH-427 use Spring bean names as IDs for lambda handlers

### LINKED ISSUE #427: Use Spring bean names as stable IDs for lambda handlers
Lambda handler class names are generated by the JVM and may change between application restarts. This makes the current class-and-method-based handler ID unsuitable for persisted outbox records.
Detect interface-based lambda handlers and use their Spring bean name as the handler ID. 
Regular handlers should retain their existing generated ID, and generated lambda class names must not be registered as legacy aliases.

Acceptance criteria:
* Lambda handlers use their Spring bean name as handler ID.
* Regular interface and annotation handlers remain unchanged.
* Legacy aliases continue to work for Spring AOP proxies.
* Generated lambda class IDs are not registered as aliases.

rolandbeisel:
Detect interface-based lambda handlers and use their Spring bean name as the stable handler ID instead of the generated JVM lambda class name.

Restrict legacy alias registration to Spring AOP proxies and add regression coverage for lambda and regular interface handler IDs.

### DIFF
--- a/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
+++ b/namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
@@ -1,5 +1,6 @@
 package io.namastack.outbox.handler
 
+import io.namastack.outbox.handler.method.handler.OutboxHandlerMethod
 import io.namastack.outbox.handler.registry.OutboxFallbackHandlerRegistry
 import io.namastack.outbox.handler.registry.OutboxHandlerRegistry
 import io.namastack.outbox.handler.scanner.HandlerScanResult
@@ -9,6 +10,7 @@ import io.namastack.outbox.handler.scanner.handler.InterfaceHandlerScanner
 import io.namastack.outbox.retry.OutboxRetryPolicyRegistry
 import org.springframework.aop.support.AopUtils
 import org.springframework.beans.factory.config.BeanPostProcessor
+import org.springframework.util.ClassUtils
 
 /**
  * Spring BeanPostProcessor that discovers and registers handlers with their fallbacks.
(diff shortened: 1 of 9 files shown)


Alek96 (approved):
The PR looks good overall, and I’m going to approve it.

There’s just one thing I have some concerns about: `BaseHandlerMethod` can now have its `id` set after the object has been created. I liked the previous approach better, where the object was immutable.

What do you think about making `Id` part of the constructor instead? Then we could have something like:

```kotlin
abstract class BaseHandlerMethod(
    val bean: Any,
    val method: Method,
    id: String? = null,
) {
    val id: String = id ?: buildId()
    ...
}

    bean: Any,
    method: Method,
    id: String? = null,
) : BaseHandlerMethod(bean, method, id)

class TypedHandlerMethod(
    bean: Any,
    method: Method,
    id: String? = null,
) : OutboxHandlerMethod(bean, method, id) {
   ...
}

class TypedHandlerMethodFactory : OutboxHandlerMethodFactory {
    override fun create(
        bean: Any,
        method: Method,
    ): OutboxHandlerMethod = TypedHandlerMethod(bean, method)

    fun createFromInterface(bean: OutboxTypedHandler<*>, handlerId: String?): TypedHandlerMethod {
        val method = ReflectionUtils.findMethod(bean, "handle", 2)
        return TypedHandlerMethod(bean, method, handlerId)
    }
}

class InterfaceHandlerScanner : HandlerScanner {
    private val typedHandlerFactory = TypedHandlerMethodFactory()
    private val genericHandlerFactory = GenericHandlerMethodFactory()
    private val typedFallbackFactory = TypedFallbackHandlerMethodFactory()
    private val genericFallbackFactory = GenericFallbackHandlerMethodFactory()

    override fun scan(
        bean: Any,
        beanName: String,
    ): List<HandlerScanResult> {
        if (bean !is OutboxTypedHandler<*> && bean !is OutboxHandler) return emptyList()

        val results = mutableListOf<HandlerScanResult>()
        val handlerId = getHandlerIdOverride(bean, beanName)

        // Register typed handler if bean implements OutboxTypedHandler<T>
        if (bean is OutboxTypedHandler<*>) {
            val handler = typedHandlerFactory.createFromInterface(bean, handlerId)

            val fallback =
                if (bean is OutboxTypedHandlerWithFallback<*>) {
                    typedFallbackFactory.createFromInterface(bean)
                } else {
                    null
                }
            results += HandlerScanResult(handler, fallback)
        }

        // Register generic handler if bean implements OutboxHandler
        if (bean is OutboxHandler) {
            val handler = genericHandlerFactory.createFromInterface(bean, handlerId)

            val fallback =
                if (bean is OutboxHandlerWithFallback) {
                    genericFallbackFactory.createFromInterface(bean)
                } else {
                    null
                }
            results += HandlerScanResult(handler, fallback)
        }

        return results
    }

    private fun getHandlerIdOverride(
        bean: Any,
        beanName: String,
    ): String? = beanName.takeIf { ClassUtils.isLambdaClass(ReflectionUtils.getTargetClass(bean)) }
}
```

PR #442, opened 2026-08-25
GH-420 Add stable handler IDs and refactor handler registration

### LINKED ISSUE #420: Stable handler identifiers
## Problem

Namastack persists a generated handler ID with every outbox record. Because the ID contains the
implementation class and method signature, refactoring or replacing a generated integration handler
can leave existing records without a matching handler. This affects the Spring Modulith integration
described in [spring-projects/spring-modulith#1777](https://github.com/spring-projects/spring-modulith/issues/1777).

## Proposal

Allow handlers to opt into an implementation-independent ID:

- `@OutboxHandler(id = "...")` for annotation-based handlers;
- override `getHandlerId()` on `OutboxHandler` or `OutboxTypedHandler` for interface-based handlers;
- keep the generated ID when no explicit ID is supplied.

Explicit IDs must be non-blank and unique. Duplicate IDs must fail application startup without
partially registering the conflicting handler. Previously generated IDs remain available as legacy
aliases so outstanding records can still be processed.

rolandbeisel:
- expose opt-in canonical IDs and lookup aliases for annotation and interface handlers
- replace scanner and factory hierarchies with focused discovery, assembly, identity, selection, and invocation components
- validate and install canonical and alias routes atomically
- preserve persisted-record, proxy, lambda, fallback, and retry-policy compatibility
- centralize handler context creation and keep default retry-policy resolution lazy
- expand Java and Kotlin regression tests and improve coverage


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/InterfaceHandlerDiscoverer.kt
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Discovers primary methods implemented through the public handler interfaces.
+ *
+ * Interface configuration, including stable IDs, aliases, and generic scheduling support, is
+ * copied into the resulting candidates for later assembly.
+ *
+ * @author Roland Beisel
+ * @since 1.8.1
+ */
+internal object InterfaceHandlerDiscoverer {
```
Alek96, line 18: Note: Make sure there's test coverage for all bean implementation possibilities: a bean that implements neither interface, a bean that implements only `OutboxTypedHandler<*>`, a bean that implements only `OutboxHandler`, and a bean that implements both at the same time (which should yield two candidates).

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/InterfaceHandlerDiscoverer.kt
```diff
@@ -0,0 +1,54 @@
...
+    /** Returns typed and generic interface candidates declared by [bean]. */
+    fun discover(
+        bean: Any,
+        beanName: String,
+    ): List<HandlerCandidate> {
+        val lambdaId = beanName.takeIf { ClassUtils.isLambdaClass(ReflectionUtils.getTargetClass(bean)) }
+        return buildList {
+            if (bean is OutboxTypedHandler<*>) {
+                add(
+                    HandlerCandidate(
```
Alek96, line 39: I'd suggest constructing `HandlerCandidate` with named arguments here. It improves readability across the 8 parameters, and lets the typed branch pass its always-true predicate explicitly as `supportsScheduling = { _, _ -> true }` instead of a trailing lambda, making the intent clearer.

```kotlin
add(
    HandlerCandidate(
        beanName = beanName,
        bean = bean,
        method = ReflectionUtils.findMethod(bean, "handle", 2),
        source = HandlerSource.TYPED_INTERFACE,
        configuredId = bean.getHandlerId(),
        configuredAliases = bean.getHandlerAliases(),
        lambdaBeanNameId = lambdaId,
        supportsScheduling = { _, _ -> true },
    ),
)
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/HandlerCandidate.kt
```diff
@@ -0,0 +1,33 @@
...
+ */
+internal data class HandlerCandidate(
+    val beanName: String,
+    val bean: Any,
+    val method: Method,
+    val source: HandlerSource,
+    val configuredId: String?,
+    val configuredAliases: Set<String>,
+    val lambdaBeanNameId: String?,
+    val supportsScheduling: (Any, OutboxRecordMetadata) -> Boolean,
```
Alek96: The name `supportsScheduling` reads like a capability flag ("does this handler support scheduling at all?"), but its actual purpose is to decide whether this specific payload is supported by the handler (e.g. only records implementing `KafkaMessage`), returning false to skip record creation — as described in `OutboxHandler.supports`. 

Potential name suggestion: 
- `supportsPayload` — it better reflects the per-record/per-payload semantics.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/HandlerDiscoveryValidator.kt
```diff
@@ -0,0 +1,44 @@
...
+            .firstOrNull { candidates -> candidates.map { it.source }.distinct().size > 1 }
+            ?.let { candidates ->
+                throw IllegalStateException(
+                    "Handler method ${candidates.first().method.toGenericString()} was discovered through both annotation and interface declarations",
+                )
+            }
+    }
+
+    /** Returns whether a primary declaration has a supported method signature. */
+    fun supportsHandler(candidate: HandlerCandidate): Boolean {
```
Alek96, line 34: This check is effectively source-specific: the TYPED_INTERFACE / GENERIC_INTERFACE branches just return true, and the only real logic (the parameter-signature validation) applies solely to HandlerSource.ANNOTATION. That makes the method read oddly.

I propose two options:
- Move it into AnnotatedHandlerDiscoverer, where the annotated signature constraints are already understood, so the annotation-specific rules live in one place.
- Keep it in the validator, accepting the tradeoff.

Which option do you prefer?

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/identity/OutboxHandlerMethodIdentityResolver.kt
```diff
@@ -0,0 +1,51 @@
...
+
+        val legacyGeneratedId = OutboxHandlerMethod.generatedId(candidate.bean, candidate.method)
+        val legacyRuntimeId =
+            OutboxHandlerMethod.generatedId(
+                candidate.bean,
+                candidate.method,
+                candidate.bean::class.java,
+            )
+        val canonicalId = candidate.configuredId ?: candidate.lambdaBeanNameId ?: legacyGeneratedId
+        val aliases =
```
Alek96, line 42: The alias-building block seems a little unclear to me. I tried to refactor it and ended up with something like this:

```kotlin
val aliases =
    buildSet {
        val isLambda = candidate.lambdaBeanNameId != null

        // user-defined aliases
        addAll(candidate.configuredAliases)

        // legacy ID generated from the unproxied target class
        if (!isLambda && legacyGeneratedId != canonicalId) {
            add(legacyGeneratedId)
        }

        // legacy ID generated from the runtime (proxy) class
        if (!isLambda && legacyRuntimeId != canonicalId && legacyRuntimeId != legacyGeneratedId) {
            add(legacyRuntimeId)
        }

        // guards against a user-supplied alias colliding with the canonical ID
        remove(canonicalId)
    }
```

Why this is equivalent to the current code:
- legacyGeneratedId branch: the original guards with configuredId != null && lambdaBeanNameId == null. Whenever configuredId != null, canonicalId equals configuredId, so legacyGeneratedId != canonicalId is already true — and the only non-lambda case where legacyGeneratedId == canonicalId is exactly when configuredId == null (canonical falls through to legacyGeneratedId). So legacyGeneratedId != canonicalId && !isLambda selects the same set of cases as configuredId != null && !isLambda.
- legacyRuntimeId branch: the original guards with isAopProxy(bean) && !isLambda && canonicalId != legacyRuntimeId. legacyRuntimeId and legacyGeneratedId differ only when the bean is AOP-proxied (proxy class name vs. target class name), so legacyRuntimeId != legacyGeneratedId is equivalent to isAopProxy(bean) here — this replaces the explicit proxy check. The !isLambda and != canonicalId conditions are unchanged. I also swapped AopUtils.getTargetClass(...)-based lambda detection for candidate.lambdaBeanNameId != null, which is derived from the same isLambdaClass(getTargetClass(...)) test upstream.
- remove(canonicalId): kept, since it's still needed to strip the canonical ID if a user happens to list it among their configured aliases (the per-branch != canonicalId guards only cover the two generated IDs).

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/identity/OutboxHandlerMethodIdentityResolver.kt
```diff
@@ -0,0 +1,51 @@
...
+    /** Resolves and validates the canonical ID and aliases for [candidate]. */
+    fun resolve(candidate: HandlerCandidate): HandlerIdentity {
+        check(candidate.configuredId == null || candidate.configuredId.isNotBlank()) {
+            "Blank handler ID configured for ${candidate.method.toGenericString()}"
+        }
+        check(candidate.configuredAliases.none { it.isBlank() }) {
+            "Blank handler alias configured for ${candidate.method.toGenericString()}"
+        }
+
+        val legacyGeneratedId = OutboxHandlerMethod.generatedId(candidate.bean, candidate.method)
```
Alek96: I'm not convinced by the name `legacyGeneratedId`. There's nothing inherently "legacy" about it — it's simply the ID generated from the target class and method signature, and when no `configuredId`/`lambdaBeanNameId` is present it actually becomes the canonical ID written to new records. Calling it "legacy" wrongly implies it's always an outdated/fallback value. I'd suggest `generatedId` instead. One thing to consider: it's currently paired with `legacyRuntimeId`, so if we drop "legacy" here it'd be worth renaming that counterpart too (e.g. `runtimeGeneratedId`) so the two stay visibly related.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/handler/OutboxHandlerMethod.kt
```diff
@@ -10,11 +12,44 @@ import java.lang.reflect.Method
...
  */
 sealed class OutboxHandlerMethod(
     bean: Any,
     method: Method,
-) : BaseHandlerMethod(bean, method)
+    canonicalId: String? = null,
+    routingAliases: Set<String> = emptySet(),
+) : InvocableHandlerMethod(bean, method) {
+    /** ID generated from the unproxied target class and reflected method signature. */
+    val legacyGeneratedId: String = generatedId(bean, method)
```
Alek96: Regardless of any user-facing benefit of exposing these two properties, I'd remove them to keep the class cleaner and simpler. No production code reads them — only tests — and `id` + `aliases` already cover the routing contract. The tests can assert via `aliases` or the `generatedId(...)` helper instead.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/handler/OutboxHandlerMethod.kt
```diff
@@ -10,11 +12,44 @@ import java.lang.reflect.Method
...
+ * @param routingAliases Additional IDs that route persisted records to this handler
  *
  * @author Roland Beisel
  * @since 1.0.0
  */
 sealed class OutboxHandlerMethod(
     bean: Any,
     method: Method,
-) : BaseHandlerMethod(bean, method)
+    canonicalId: String? = null,
```
Alek96: `canonicalId` can be a non-nullable `String` here, as we're always going to provide an explicit value: `OutboxHandlerMethodIdentityResolver` computes it via 
```kotlin
val canonicalId = candidate.configuredId ?: candidate.lambdaBeanNameId ?: legacyGeneratedId
```
so `HandlerIdentity.canonicalId` is never `null` and `HandlerMethodFactory` always passes a concrete value in.

The same comment applies to `GenericHandlerMethod` and `TypedHandlerMethod`.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/assembly/HandlerRetryPolicyResolver.kt
```diff
@@ -0,0 +1,33 @@
...
+    private val registry: OutboxRetryPolicyRegistry,
+) {
+    /** Resolves annotation-based policies before an [OutboxRetryAware] policy. */
+    fun resolve(candidate: HandlerCandidate): OutboxRetryPolicy? {
+        AnnotatedElementUtils.findMergedAnnotation(candidate.method, OutboxRetryable::class.java)?.let { annotation ->
+            if (annotation.value != OutboxRetryPolicy::class) return registry.getRetryPolicy(annotation.value)
+            if (annotation.name.isNotBlank()) return registry.getRetryPolicy(annotation.name)
+        }
+
+        return (candidate.bean as? OutboxRetryAware)?.getRetryPolicy()
```
Alek96, line 39: The `OutboxRetryAware` fallback (line 31) matches on the bean without checking `candidate.source`, so it also applies to annotation-based handlers. Note this preserves main's branch behavior — `RetryPolicyScanner.scan` did the same `if (bean is OutboxRetryAware) return bean.getRetryPolicy()` with no source filtering — so it's not a regression introduced here. 

That said, since `OutboxRetryAware.getRetryPolicy()` is a bean-level signal, on a bean with multiple `@OutboxHandler` methods it applies one policy to every method, and the docs only ever illustrate it alongside interface handlers. If the intent is for `OutboxRetryAware` to be interface-only (`OutboxHandler`/`OutboxTypedHandler`), this PR would be a good place to gate it via `candidate.source` and add a test — but that's a behavior change to decide on deliberately, not a fix required by the refactoring.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
```diff
@@ -114,7 +114,13 @@ class OutboxHandlerRegistry {
      * @param handlerMethod The handler method to register
      * @throws IllegalStateException if a handler with the same ID already exists
      */
-    internal fun register(handlerMethod: OutboxHandlerMethod) {
+    internal fun register(handlerMethod: OutboxHandlerMethod) = register(handlerMethod, "<unknown>")
```
Alek96: Is there any purpose in keeping the `register` / `registerAlias` methods? Production goes exclusively through registerBatch, and these are only called from tests. If not, we can remove them — they're internal, so users never relied on them.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
```diff
@@ -28,6 +28,8 @@ class OutboxHandlerRegistry {
      * Used for direct handler lookup via metadata.handlerId.
      */
     private val handlersById = mutableMapOf<String, OutboxHandlerMethod>()
+    private val registrationsById = mutableMapOf<String, HandlerRegistration>()
+    private val routingOwners = mutableMapOf<String, RoutingOwner>()
```
Alek96: `routingOwners` is only populated and read during registration (startup) for collision messages; after startup it's never used but stays in memory for the registry's lifetime. Since `handlersById` / `registrationsById` already map every routing ID to its handler and registration, I'd suggest replacing the stored map with a private method that returns the `RoutingOwner` for a given routing ID on demand, so we don't keep a third map around just for validation.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
```diff
@@ -67,61 +45,35 @@ internal class OutboxHandlerBeanPostProcessor(
...
+        if (registrations.isNotEmpty()) handlerRegistry.registerBatch(registrations)
 
-            // a. Register handler
-            handlerRegistry.register(handler)
+        registrations.forEach { registration ->
+            val handler = registration.primary
 
-            // b. Register fallback if present (1:1 mapping)
-            result.fallback?.let { fallback ->
+            // Compatibility projections for processors that consume the legacy facades.
```
Alek96: Do you plan to migrate the legacy facades (`OutboxFallbackHandlerRegistry` / `OutboxRetryPolicyRegistry`) onto the new `OutboxHandlerRegistry`?

Alek96 (commented):
I like what you did with the bean handler discovery and registration. Nice job 👍

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/InterfaceHandlerDiscoverer.kt
rolandbeisel: Added explicit test coverage for beans implementing neither interface, only OutboxTypedHandler, only OutboxHandler, and both interfaces. The combined case verifies that two candidates are discovered.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/InterfaceHandlerDiscoverer.kt
rolandbeisel: Applied. HandlerCandidate is now constructed with named arguments, and the always-true payload predicate is passed explicitly for typed handlers.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/HandlerCandidate.kt
rolandbeisel: Renamed consistently to `supportsPayload` across `HandlerCandidate`, `GenericHandlerMethod`, the selection logic, tests, and KDoc. There are no remaining supportsScheduling references.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/HandlerDiscoveryValidator.kt
rolandbeisel: I prefer to keep this check in HandlerDiscoveryValidator. Although the current signature-specific branch only applies to annotation handlers, keeping discovery and validation separate gives each component a clear responsibility and provides a central place for future declaration constraints.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/identity/OutboxHandlerMethodIdentityResolver.kt
rolandbeisel: Applied the suggested simplification. Alias construction now uses the relationships between the generated IDs directly and derives the lambda case from lambdaBeanNameId.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/method/handler/OutboxHandlerMethod.kt
rolandbeisel: Agreed. These handler method classes are implementation details and are not intended for direct user construction. canonicalId is now required and non-nullable across OutboxHandlerMethod, GenericHandlerMethod, and TypedHandlerMethod; generated identity remains the responsibility of the identity resolver.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/assembly/HandlerRetryPolicyResolver.kt
rolandbeisel: I verified this against main. RetryPolicyScanner applied OutboxRetryAware without checking the handler source, so annotation-based handlers already inherited the bean-level policy. The refactored resolver preserves the same priority: method-level @OutboxRetryable, then OutboxRetryAware, then the default policy. The only operational difference is that the policy is now resolved once per handler declaration and reused for canonical and alias routes instead of potentially invoking getRetryPolicy() again for a proxy alias. I kept the existing source-independent behavior and added regression tests for annotation-based handlers and annotation precedence.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
rolandbeisel: Applied. The test-only registration methods were removed, and the tests now exercise the same atomic registerBatch path used by production.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
rolandbeisel: Applied. Routing ownership is now reconstructed on demand from the existing registration data, so the additional lifetime map is no longer retained.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
rolandbeisel: Not as part of this PR. Consolidating handler-related fallback and retry-policy mappings into OutboxHandlerRegistry could reduce duplicated state and alias projections, but it affects processors, invokers, auto-configuration, and existing public registry types. We may revisit this as a dedicated follow-up change.

Alek96: You don’t necessarily have to remove either the fallback registry or the retry registry. As a first step, we could make both of them use `OutboxHandlerRegistry`, so this wouldn’t affect processors, invokers, auto-configuration, or the existing public registry types. At the same time, we wouldn’t need to double-register everything.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/assembly/HandlerMethodFactory.kt
```diff
@@ -0,0 +1,48 @@
...
+ * @author Roland Beisel
+ * @since 1.8.1
+ */
+internal object HandlerMethodFactory {
+    /** Creates a typed or generic primary method according to the declared payload type. */
+    fun primary(
+        candidate: HandlerCandidate,
+        identity: HandlerIdentity,
+    ): OutboxHandlerMethod =
+        if (candidate.method.parameterTypes.first() == Any::class.java) {
```
Alek96: Now that I look at it more closely, I think there is a subtle case where this won't work. If a class implements `OutboxTypedHandler<Any>`, it will be categorized as `Generic` instead of `Typed`.

Interestingly, even if the `Typed` interface is categorized as `Generic`, the code should still work without any issues.

You can decide whether to keep it as is or fix the `if` condition.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/HandlerDiscoveryValidator.kt
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Validates declaration relationships and supported signatures before handler assembly.
+ *
+ * @author Roland Beisel
+ * @since 1.8.1
+ */
+internal object HandlerDiscoveryValidator {
+    /** Rejects a method declared through both an annotation and a handler interface. */
+    fun validateRelationships(declarations: HandlerDeclarations) {
```
Alek96, line 20: The doc comment and exception message state this rejects a method declared through both an annotation and an interface, but the grouping logic (`groupBy { it.method }` → distinct source count > 1) is actually broader. If I'm reading it right, a bean that implements both `OutboxHandler` and `OutboxTypedHandler<Any>` resolves both branches to the same erased `handle(Any, OutboxRecordMetadata)` method, giving one Method mapped to two distinct sources (`TYPED_INTERFACE` + `GENERIC_INTERFACE`) — which would also trip this rejection.

If I am right, we could split this validation into two — first checking this edge case with a proper, dedicated message, then the double-source (annotation + interface) case. Extracting both into private methods (as plain filters, no groupBy) would also make the intent clearer:

```kotlin
fun validateRelationships(declarations: HandlerDeclarations) {
    rejectAmbiguousInterfaceHandlers(declarations.handlers)
    rejectMixedAnnotationAndInterface(declarations.handlers)
}

private fun rejectAmbiguousInterfaceHandlers(candidates: List<HandlerCandidate>) {
    val typed = candidates.filter { it.source == HandlerSource.TYPED_INTERFACE }.map { it.method }.toSet()
    candidates
        .groupBy { it.method }
        .values
        .firstOrNull { candidates ->
            val sources = candidates.map { it.source }.toSet()
            HandlerSource.TYPED_INTERFACE in sources && HandlerSource.GENERIC_INTERFACE in sources
        }
        ?.let { candidates ->
            throw IllegalStateException(
                "Handler method ${candidates.first().method.toGenericString()} implements both OutboxHandler and " +
                    "OutboxTypedHandler<Any>, resulting in an ambiguous handle(Any, OutboxRecordMetadata) signature",
            )
        }
}

private fun rejectMixedAnnotationAndInterface(candidates: List<HandlerCandidate>) {
    candidates
        .groupBy { it.method }
        .values
        .firstOrNull { group -> group.map { it.source }.toSet().let { HandlerSource.ANNOTATION in it && it.size > 1 } }
        ?.let { group ->
            throw IllegalStateException(
                "Handler method ${group.first().method.toGenericString()} was discovered through both annotation and interface declarations",
            )
        }
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/invoker/OutboxHandlerInvoker.kt
```diff
@@ -49,15 +47,13 @@ class OutboxHandlerInvoker(
      */
     fun dispatch(record: OutboxRecord<*>) {
         val payload = record.payload ?: return
-        val metadata = record.toMetadata()
+        val metadata = OutboxHandlerContextFactory.metadata(record)
 
         val handler =
-            handlerRegistry.getHandlerById(record.handlerId)
+            handlerRegistry.getRegistrationById(record.handlerId)?.primary
```
Alek96: Why do we have both?

`handlerRegistry.getRegistrationById(record.handlerId)?.primary` and `handlerRegistry.getHandlerById(record.handlerId)`?

`handlerRegistry.getHandlerById` should be sufficient here.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
```diff
@@ -82,7 +86,7 @@ class OutboxHandlerRegistry {
      * @return List of TypedHandlerMethods for this type (empty if none)
      */
     fun getHandlersForPayloadType(type: KClass<*>): List<TypedHandlerMethod> =
-        typedHandlers[type]?.toList() ?: emptyList()
+        OutboxHandlerSelector.typed(typedHandlers, type)
```
Alek96: I'm not sure we need this separate class. Both methods are one-liners — `typed` is just `handlers[type]?.toList() ?: emptyList()`, and `generic` is a single filter on the scheduling predicate — and each has exactly one caller in `OutboxHandlerRegistry`. 

Since the registry already owns `typedHandlers` and `genericHandlers`, I'd suggest inlining these two expressions into `getHandlersForPayloadType` / `getGenericHandlers` and dropping the class. The only tradeoff is losing it as an independently-named "selection" unit, but with single-expression, single-caller methods that separation doesn't buy much.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/RoutingOwner.kt
```diff
@@ -0,0 +1,22 @@
...
+ * handler together makes startup collision errors identify both conflicting declarations.
+ *
+ * @property role whether the route is a canonical ID or an alias
+ * @property beanName Spring bean that contributed the route
+ * @property handler primary handler reached through the route
+ *
+ * @author Roland Beisel
+ * @since 1.8.1
+ */
+internal data class RoutingOwner(
```
Alek96, line 18: `RoutingOwner` is an internal data class used only inside `OutboxHandlerRegistry`. Since it's purely a helper for the registry's collision detection with no other consumers, I'd suggest moving it into `OutboxHandlerRegistry` as a private nested class and dropping the standalone file. That tightens encapsulation and keeps it out of the package namespace.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerBeanPostProcessor.kt
rolandbeisel: Agreed, I implemented this approach. OutboxHandlerRegistry is now the single source for complete handler registrations, including fallbacks, explicit retry policies, canonical IDs, and aliases. OutboxFallbackHandlerRegistry and OutboxRetryPolicyRegistry remain as compatibility facades for the existing processor and invoker boundaries, but they read handler-specific data directly from OutboxHandlerRegistry and no longer maintain duplicate maps. The BeanPostProcessor now performs only the atomic batch registration. The default and bean-based retry-policy resolution remains owned by OutboxRetryPolicyRegistry.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/assembly/HandlerMethodFactory.kt
rolandbeisel: Fixed. `HandlerMethodFactory` now classifies interface handlers based on `HandlerSource` instead of the erased payload parameter type. Therefore, `OutboxTypedHandler<Any>` remains a `TypedHandlerMethod`, while annotation-based handlers are still classified by their declared parameter type. I also added a regression test for this case.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/discovery/HandlerDiscoveryValidator.kt
rolandbeisel: Good catch. I split the validation into two dedicated checks: one for the ambiguous `OutboxHandler` + `OutboxTypedHandler<Any>` combination and one for methods discovered through both annotation and interface declarations. Both cases now have distinct error messages and are covered by unit and startup integration tests.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/invoker/OutboxHandlerInvoker.kt
rolandbeisel: Agreed. OutboxHandlerInvoker now uses `handlerRegistry.getHandlerById(record.handlerId)` directly. The complete registration lookup remains internal and is only used where fallback or retry-policy information is actually required.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/OutboxHandlerRegistry.kt
rolandbeisel: Applied. The typed lookup and generic payload filtering are now implemented directly in OutboxHandlerRegistry, and the separate OutboxHandlerSelector class has been removed.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/registry/RoutingOwner.kt
rolandbeisel: I decided to keep RoutingOwner as a separate internal class. The handler package follows a one-class-per-file structure, and keeping this collision-diagnostic concept in its own file is more consistent with that organization. It remains internal and is only used by OutboxHandlerRegistry.

Alek96 (dismissed):

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerIdentity.kt
```diff
@@ -0,0 +1,19 @@
...
+ * handler identity while still allowing [aliases] to be configured. Aliases are lookup-only
+ * identities used to route records persisted under an earlier handler identity.
+ *
+ * @property id Optional stable canonical handler ID; `null` retains the generated identity
+ * @property aliases Alternative lookup-only IDs for persisted records
+ *
+ * @author Roland Beisel
+ * @since 1.8.1
+ */
+data class OutboxHandlerIdentity(
```
Alek96, line 16: Consider adding `@JvmOverloads` so Kotlin generates actual Java overload constructors:

```java
new OutboxHandlerIdentity()
new OutboxHandlerIdentity("id")
new OutboxHandlerIdentity("id", aliases)
```

Without `@JvmOverloads`, Java will see only one constructor:

```java
new OutboxHandlerIdentity("id", aliases)
```

so callers must explicitly provide both arguments.

Alek96 (approved):

namastack-outbox-api/src/main/kotlin/io/namastack/outbox/handler/OutboxHandlerIdentity.kt
rolandbeisel: Good point. I added `@JvmOverloads`, so Java callers can now use the no-arg, ID-only, or full constructor. I also added a Java compatibility test and updated the documentation examples.

PR #450, opened 2026-08-31
1. Instrumentation SPI

rolandbeisel:


namastack-outbox-observability/src/main/kotlin/io/namastack/outbox/observability/OutboxObservabilityAutoConfiguration.kt
```diff
@@ -35,122 +25,38 @@ import org.springframework.context.annotation.Role
...
+    @ConditionalOnMissingBean(MicrometerOutboxInstrumentation::class)
+    fun micrometerOutboxInstrumentation(
+        observationRegistry: ObservationRegistry,
+        scheduleConvention: ObjectProvider<OutboxScheduleObservationConvention>,
+        processConvention: ObjectProvider<OutboxProcessObservationConvention>,
+    ): MicrometerOutboxInstrumentation =
+        MicrometerOutboxInstrumentation(
+            observationRegistry = observationRegistry,
+            customScheduleConvention = scheduleConvention.getIfAvailable(),
+            customProcessConvention = processConvention.getIfAvailable(),
```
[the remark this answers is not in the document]
rolandbeisel: @codex should be fixed now

namastack-outbox-observability/src/main/kotlin/io/namastack/outbox/observability/OutboxObservabilityAutoConfiguration.kt
```diff
@@ -35,122 +25,38 @@ import org.springframework.context.annotation.Role
...
-                        channelNameProvider.getIfAvailable { OutboxChannelNameProvider.DEFAULT }
-                    },
-                )
-            return DefaultPointcutAdvisor(pointcut, advice)
-        }
-    }
+    @ConditionalOnBean(ObservationRegistry::class)
+    @ConditionalOnMissingBean(MicrometerOutboxInstrumentation::class)
+    fun micrometerOutboxInstrumentation(
+        observationRegistry: ObservationRegistry,
```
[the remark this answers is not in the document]
rolandbeisel: @codex can you review again?

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/context/OutboxContextCollector.kt
```diff
@@ -63,15 +64,23 @@ import org.slf4j.LoggerFactory
...
  * @author Aleksander Zamojski
  * @since 1.0.0
  */
-class OutboxContextCollector(
-    private val providers: List<OutboxContextProvider>,
+class OutboxContextCollector internal constructor(
+    providersSupplier: () -> List<OutboxContextProvider>,
 ) {
     private val log = LoggerFactory.getLogger(OutboxContextCollector::class.java)
+    private val providers: List<OutboxContextProvider> by lazy(SYNCHRONIZED) {
```
Alek96: Nit: `lazy(SYNCHRONIZED)` is redundant — prefer plain `lazy { }`

`LazyThreadSafetyMode.SYNCHRONIZED` is already the default mode for `lazy {}`, so `lazy(SYNCHRONIZED, supplier)` and `lazy(SYNCHRONIZED) { ... }` are behaviorally identical to `lazy(supplier)` / `lazy { ... }`. The explicit argument adds an extra `import kotlin.LazyThreadSafetyMode.SYNCHRONIZED` in each file without changing anything.

It's also inconsistent with the rest of the codebase, which uses plain `lazy {}` everywhere (`PartitionCoordinator`, `OutboxRetryPolicyRegistry`, `MongoCollectionNameResolver`, `DefaultJdbcTableNameResolver`). The explicit form only appears in the files this PR touches.

Thread-safety here is legitimate (these resolve on the first outbox operation, which can be reached concurrently), but plain `lazy {}` already guarantees exactly that — so dropping the explicit mode keeps the same guarantee.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/config/OutboxCoreInfrastructureAutoConfiguration.kt
```diff
@@ -38,31 +40,55 @@ import java.time.Clock
...
+        instrumentations: ObjectProvider<OutboxInstrumentation>,
+        channelNameProvider: ObjectProvider<OutboxChannelNameProvider>,
+    ): OutboxHandlerInvoker =
         OutboxHandlerInvoker(
             handlerRegistry = outboxHandlerRegistry,
+            instrumentationSupplier = {
+                OutboxInstrumentation.compose(instrumentations.orderedStream().toList())
+            },
+            channelNameProviderSupplier = {
+                channelNameProvider.getIfAvailable { OutboxChannelNameProvider.DEFAULT }
```
Alek96: Question / nit: unreachable `{ DEFAULT }` fallback in the channel-provider suppliers

The `getIfAvailable { OutboxChannelNameProvider.DEFAULT }` fallback looks redundant. Since this same config declares:
```kotlin
@Bean
@ConditionalOnMissingBean
fun outboxChannelNameProvider(): OutboxChannelNameProvider = OutboxChannelNameProvider.DEFAULT
```
an `OutboxChannelNameProvider` bean is always present, so the `ObjectProvider` will always resolve and the `{ DEFAULT }` lambda is effectively dead code — it can never fire.

I suggest simplifying the supplier body to `channelNameProvider.getObject()` — same behavior, but fail-fast if the bean is ever genuinely missing (which would signal a real misconfiguration rather than silently defaulting), and it removes a fallback that reads as meaningful but can never execute.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxService.kt
```diff
@@ -450,11 +438,58 @@ class OutboxService(
     override fun schedule(payload: Any) {
```
Alek96, line 437: I'm not entirely sure about the current solution here. `schedule(payload)` now delegates to the `schedule(payload, additionalContext)` overload, which then routes to `scheduleInstrumented`, whereas the other overloads call `scheduleInstrumented` directly. Honestly, I think I preferred the old way where we duplicated a bit of code — each overload was self-contained, and you could see the auto-generated-key intent right at the entry point, rather than it being implicitly inherited one level down.

Something like this, where the overload calls scheduleInstrumented directly with the explicit auto-generated key:
```kotlin
override fun schedule(payload: Any) {
    scheduleInstrumented(
        payload = payload,
        key = UUID.randomUUID().toString(),
        additionalContext = emptyMap(),
        instrumentationKey = "auto-generated",
    )
}
```

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instrumentation/OutboxInstrumentation.kt
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Instruments scheduling and processing at the stable outbox operation boundaries.
+ *
+ * Implementations are observational around-interceptors. They must invoke action exactly once
+ * and propagate its result or failure unchanged.
+ *
+ * @author Roland Beisel
+ * @since 1.10.0
+ */
+interface OutboxInstrumentation {
```
Alek96, line 12: Did you consider splitting `OutboxInstrumentation` into schedule and process?

Just flagging this for discussion — not asking to change anything. `OutboxInstrumentation` is currently one interface with two methods (`schedule` and `process`). I'm wondering whether you considered splitting it and want to make sure the trade-off was deliberate, since this is a public SPI and the choice is sticky once released.

**Option A — keep it as one interface (current)**
```kotlin
interface OutboxInstrumentation {
    fun schedule(invocation: OutboxScheduleInvocation, action: () -> Unit)
    fun process(invocation: OutboxProcessInvocation, action: () -> Unit)
}
```

For:
- Conceptual cohesion — both methods instrument "outbox operation boundaries" and share the same lifecycle, ordering, and composition machinery.
- Every current implementation (`MicrometerOutboxInstrumentation`, the logging example) instruments both boundaries anyway.
- Simplest wiring — Core injects one `ObjectProvider<OutboxInstrumentation>`, composes once, and threads a single instance into `OutboxService`, `OutboxHandlerInvoker`, and `OutboxFallbackHandlerInvoker`. One `CompositeOutboxInstrumentation`, one `NOOP`.

Against:
- Mild ISP violation: an implementer that only cares about scheduling still has to implement the `process` (even if it just calls `action()`), and vice versa.
- `OutboxService` never calls `process`, yet it depends on an interface that carries it — the dependency is slightly wider than the actual usage.

**Option B — split into two interfaces + a convenience combined interface**
```kotlin
interface OutboxScheduleInstrumentation {
    fun schedule(invocation: OutboxScheduleInvocation, action: () -> Unit)
}
interface OutboxProcessInstrumentation {
    fun process(invocation: OutboxProcessInvocation, action: () -> Unit)
}
// convenience for implementers that want both
interface OutboxInstrumentation :
    OutboxScheduleInstrumentation,
    OutboxProcessInstrumentation
```

For:
- ISP-clean — implement only the boundary you care about (a schedule-only auditor implements one method).
- More honest dependencies — `OutboxService` depends only on the schedule contract, the invokers only on the process contract.
- Still supports "both" via the combined convenience interface, so existing implementations (Micrometer, logging example) barely change.

Against:
- More surface area and wiring — two facets to inject/compose/no-op (two `ObjectProviders` and two composites, or a generic composite), and the auto-config grows accordingly.
- Beans implementing the combined interface need to be picked up by both facets — a bit more registration plumbing.
- Three public types to document and maintain instead of one.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instrumentation/OutboxInstrumentation.kt
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Instruments one scheduling operation.
+     *
+     * @param invocation Description of the scheduling operation.
+     * @param action Scheduling action to invoke exactly once.
+     */
+    fun schedule(
```
Alek96, line 19: Consider default no-op method bodies on OutboxInstrumentation

Independent of the single-vs-split discussion above: if we do keep the single interface, it might be worth giving the two methods default no-op bodies that just invoke `action()`.

```kotlin
interface OutboxInstrumentation {
    fun schedule(
        invocation: OutboxScheduleInvocation,
        action: () -> Unit,
    ) = action()

    fun process(
        invocation: OutboxProcessInvocation,
        action: () -> Unit,
    ) = action()

    // companion with NOOP / compose(...) unchanged
}
```

For:
- Directly addresses the mild ISP concern from the interface discussion — an implementer that only cares about scheduling can override just schedule and leave process as the pass-through default (and vice versa), instead of writing boilerplate for the boundary it doesn't care about.
- Keeps all the single-interface wiring simplicity (one ObjectProvider, one composite, one bean type).

Against:
- Slightly weakens the "instrument both boundaries" framing in the current KDoc — a default means "do nothing" becomes a silent, valid choice rather than an explicit one.


Note: if we're leaving the current implementation as a single interface, adding these default no-op bodies could be worth it — we could then delete `NoOpOutboxInstrumentation` entirely. It's only referenced in one place (backing `OutboxInstrumentation.NOOP`), and once the interface provides the pass-through defaults, `NOOP` no longer needs a dedicated class:

```kotlin
// NoOpOutboxInstrumentation.kt deleted entirely

// in OutboxInstrumentation companion — NOOP becomes a trivial empty instance
companion object {
    @JvmField
    val NOOP: OutboxInstrumentation = object : OutboxInstrumentation {}

    // compose(...) unchanged
}
```
That removes a whole file and the duplicated pass-through logic, while `OutboxInstrumentation.NOOP` (used by the constructor defaults, compose, and tests) keeps working unchanged.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/context/OutboxContextCollector.kt
Alek96, line 1: Future consideration: unify OutboxContextCollector with the compose(...) pattern used for instrumentation

Not part of this PR — just capturing an idea for a follow-up. Now that `OutboxInstrumentation` establishes a clean composition pattern (a single unit interface, a `CompositeOutboxInstrumentation`, and a `compose()` factory that collapses 0 → NOOP / 1 → single / many → composite), we could align context collection to the exact same shape:
- `OutboxContextProvider` = the unit (analogous to a single `OutboxInstrumentation`)
- `OutboxContextCollector` = the composite (analogous to `CompositeOutboxInstrumentation`), implementing `OutboxContextProvider` itself as a merging fan-out
- a `compose(...)`-style factory: 0 providers → NOOP (returns emptyMap()), 1 provider → the user's provider directly, 2+ → `OutboxContextCollector`
- `OutboxService` would then depend on `OutboxContextProvider` resolved from `ObjectProvider<OutboxContextProvider>` via that factory, with no dedicated `outboxContextCollector` bean — mirroring exactly how instrumentation is wired.

For:
- Full structural symmetry with the instrumentation pattern (`Provider` ~ `Instrumentation`, `Collector` ~ `CompositeInstrumentation`, `compose()` on both) — understanding one immediately explains the other.
- `OutboxService` depends on the interface (“give me merged context”) rather than a concrete collector.
- Removes the dedicated bean; composition happens inline via `ObjectProvider`, same as instrumentation.

Against / things to resolve:
- Identity shift: `OutboxContextCollector` would become an `OutboxContextProvider` (an aggregator that is-a source). Defensible as a "merging provider," but it blurs the "provider = source of context" mental model a little.
- Breaking public API: `OutboxContextCollector` is `@since 1.0.0` with a public `constructor(List<OutboxContextProvider>)` and `collectContext()`. Making it a provider (and/or exposing `provide()`) is a breaking change for non-Spring/direct users unless we keep the existing surface alongside.
- Divergent error contract — the key caveat: the two patterns intentionally behave oppositely. Instrumentation rethrows failures unchanged (fail-fast pass-through), which is exactly what makes its 1 → single shortcut free. Context collection deliberately swallows per-provider failures (try/catch + warn-log + continue). So we lose that fail-safety in the 1 → single and NOOP shortcuts unless we explicitly re-add it. Concretely: for the single-provider case we'd wrap it so a throwing provider is still logged-and-skipped, rather than propagating like instrumentation does. That means the collapse rule isn't a pure copy of `OutboxInstrumentation.compose(...)` — the 0/1 cases still need the resilient wrapper.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/context/OutboxContextCollector.kt
rolandbeisel: fixed

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/config/OutboxCoreInfrastructureAutoConfiguration.kt
rolandbeisel: fixed

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/OutboxService.kt
rolandbeisel: I reconsidered this, but I’d prefer to keep the delegation.
`schedule(payload, key)` follows the same pattern by delegating to the overload with `additionalContext`. This keeps both convenience overloads consistent and leaves UUID generation and the "auto-generated" instrumentation value in one place.

The instrumentation is still invoked exactly once, so the difference is only structural. For me, the consistent overload hierarchy and avoiding duplicated logic outweigh having the generated-key intent directly visible in this entry point.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instrumentation/OutboxInstrumentation.kt
rolandbeisel: Yes, I considered the split and decided to keep a single interface. Scheduling and processing are both outbox operation boundaries with the same lifecycle, ordering, and composition semantics, and the current implementations instrument both. Splitting the contract would add public types and separate composition/wiring without a concrete use case that currently requires it.

To address the concern for implementations interested in only one boundary, I added pass-through default implementations for both methods. A schedule-only or process-only instrumentation can now override just the relevant method while we retain the simpler combined SPI.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/instrumentation/OutboxInstrumentation.kt
rolandbeisel: fixed

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/context/OutboxContextCollector.kt
rolandbeisel: Thanks for capturing this. I agree that the structural symmetry is worth exploring in a separate follow-up.
I’ll leave this PR unchanged and treat it as a separate design topic. Feel free to create an issue.

Issue #449, opened 2026-08-31
Add composable instrumentation SPI to Outbox Core


rolandbeisel:
Introduce a framework-neutral OutboxInstrumentation SPI for scheduling and handler-processing operations.

Current observability relies on Spring AOP, which only applies to Spring-managed objects and can obscure the actual operation boundary. Explicit instrumentation provides consistent behavior for both Spring-managed and programmatically constructed components.

The implementation should include:

- Schedule and process invocation metadata
- Primary/fallback handler identification
- A NOOP implementation
- Ordered composition of multiple instrumentations
- No Spring, Micrometer, or Observability dependency in Core

Instrumentation must invoke actions exactly once and propagate failures unchanged.

**Acceptance criteria**

- Empty composition uses NOOP
- Multiple instrumentations execute in deterministic order
- Invocation metadata is preserved
- Exceptions propagate unchanged
- Core checks pass

PR #453, opened 2026-09-02
2. Outbox Runtime

rolandbeisel:


namastack-outbox-core/src/main/kotlin/io/namastack/outbox/config/OutboxCoreInfrastructureAutoConfiguration.kt
```diff
@@ -57,6 +55,7 @@ class OutboxCoreInfrastructureAutoConfiguration {
 
     @Bean
     @ConditionalOnMissingBean
+    @ConditionalOnProperty(name = ["namastack.outbox.mode"], havingValue = "single", matchIfMissing = true)
```
Alek96: Suggestion: extract the repeated `@ConditionalOnProperty(mode=single)` into a meta-annotation

The mode gate `@ConditionalOnProperty(name = ["namastack.outbox.mode"], havingValue = "single", matchIfMissing = true)` is now duplicated across a lot of beans and auto-configs — the Core infrastructure config alone repeats it ~11 times, plus the threading / processing / scheduling / multicaster configs, the actuator, and the JDBC / JPA / MongoDB persistence configs.

That much repetition of a string-keyed condition is a bit of a maintenance risk:

- the property name `"namastack.outbox.mode"`, the value `"single"`, and `matchIfMissing = true` all have to stay in sync by hand across every site — a single typo silently changes gating behavior;
- the *intent* ("only in single-runtime mode") is obscured by the boilerplate;
- when CHANNELS gating rules evolve, every site has to be touched individually.

I'd suggest folding it into a single composed annotation, e.g.:

```kotlin
@Target(AnnotationTarget.CLASS, AnnotationTarget.FUNCTION)
@Retention(AnnotationRetention.RUNTIME)
@ConditionalOnProperty(
    name = ["namastack.outbox.mode"],
    havingValue = "single",
    matchIfMissing = true,
)
```
annotation class ConditionalOnSingleRuntimeMode
Then each site becomes just `@ConditionalOnSingleRuntimeMode`, which reads as intent and keeps the property name/value/matchIfMissing defined in exactly one place. Spring treats `@ConditionalOnProperty` as a meta-annotation, so this composes cleanly.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/assembly/HandlerRegistrationAssembler.kt
```diff
@@ -28,31 +29,42 @@ internal class HandlerRegistrationAssembler(
...
      * @param declarations Primary and fallback declarations discovered on one bean
+     * @param primaryMethodPredicate Predicate selecting the primary methods to assemble after all
+     * declaration relationships have been validated
      * @return Complete registrations ready to be installed in the handler registry
      * @throws IllegalStateException if the declarations contain an ambiguous relationship
      */
-    fun assemble(declarations: HandlerDeclarations): List<HandlerRegistration> {
+    fun assemble(
+        declarations: HandlerDeclarations,
+        primaryMethodPredicate: (Method) -> Boolean = { true },
```
Alek96: Naming: `primaryMethodPredicate` describes the mechanism, not the purpose

The `primaryMethodPredicate: (Method) -> Boolean` parameter is threaded through `HandlerRegistrationAssembler`, `OutboxHandlerRegistrar`, and `OutboxHandlerInfrastructure`. The name describes *how* it works (a predicate over primary methods) rather than *what it's for* — selecting which handlers belong to a given runtime/channel (e.g. via a channel-selecting annotation the supporting module inspects).

A couple of things make the current name read a bit off:

- It leaks the mechanism ("predicate over methods") and hides the intent ("route this handler to the runtime it belongs to"). In `OutboxHandlerInfrastructure.register(...)` it just looks like a generic method filter with a `{ true }` default, with no signal that this is the handler-routing seam.
- "primary" is slightly misleading — filtering on the primary `method` includes/excludes the whole registration (primary + matched fallback + retry policy travel together), so it's really selecting *handlers/registrations*, not just primary methods. A reader might think fallbacks are filtered independently.
- This is the extension point a supporting module hooks into, so a purpose-oriented name would document the design intent across all three classes at once.

I don't have a strong preference on the final name — leaving the options here for you to pick whichever fits the intended layering best:

**Channel-agnostic (keeps Core neutral, no channel concept leaked):**
- `handlerSelector`
- `handlerFilter` / `includeHandler`
- `ownedHandlerPredicate` — ties into the "runtime owns a subset" model already used in this PR (`ownedResources`)

**Channel-aware (if Core is allowed to know about channel routing):**
- `channelHandlerPredicate`
- `routesToThisRuntime`

Related question rather than a naming point: should the predicate see the full discovered candidate (bean + method + annotations) instead of just the raw `Method`? If the routing decision is based on an annotation, `(HandlerCandidate) -> Boolean` might be more convenient than `(Method) -> Boolean` — though that's a separate change from the rename.

Whatever name you choose, applying it consistently across `HandlerRegistrationAssembler`, `OutboxHandlerRegistrar`, and `OutboxHandlerInfrastructure` (plus the KDoc) would make the routing intent clear.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntime.kt
```diff
@@ -0,0 +1,170 @@
...
+    @Synchronized
+    override fun close() {
+        if (closed) return
+
+        closed = true
+        val failure = cleanup()
+        if (failure != null) throw failure
+    }
+
+    private fun scheduleRebalancing(): ScheduledFuture<*> {
```
Alek96: Move rebalance scheduling into `PartitionCoordinator` (implement `SmartLifecycle`)

`OutboxRuntime.scheduleRebalancing()` currently owns the periodic rebalance scheduling: it builds a `ScheduledMethodRunnable` via reflection (`REBALANCE_METHOD`), calls `taskScheduler.scheduleWithFixedDelay(...)`, holds the `ScheduledFuture`, and cancels it during cleanup. I think this responsibility belongs in `PartitionCoordinator` instead.

A few reasons:

- **`PartitionCoordinator` already owns rebalance scheduling in the OSS/single path** — `rebalance()` carries `@Scheduled(initialDelayString = "0", fixedDelayString = "...rebalance-interval...", scheduler = "outboxDefaultScheduler")`. So we now have *two* parallel mechanisms scheduling the same operation: the annotation on the bean path, and the programmatic `scheduleWithFixedDelay` on the runtime path. That's a duplicated concern living in two places; consolidating both into the coordinator removes the split.
- **It drops the reflection hack.** If the coordinator schedules its own `rebalance()`, there's no need for `ScheduledMethodRunnable` / `REBALANCE_METHOD` in `OutboxRuntime` — it can schedule the method directly.
- **It removes rebalance-scheduling knowledge (`taskScheduler`, `rebalanceInterval`, observation registry, the `ScheduledFuture` lifecycle) from `OutboxRuntime`**, leaving `OutboxRuntime` to orchestrate ordering rather than to know *how* rebalancing is scheduled.

Concretely, I'd suggest `PartitionCoordinator` **own start and stop of its own rebalance schedule by implementing `SmartLifecycle`**, exactly like `OutboxInstanceRegistry` already does (`getPhase()` / `isRunning()` / `start()` / `stop()`, holding its own `TaskScheduler` and `ScheduledFuture`). That keeps the two lifecycle-managed components symmetric, and `OutboxRuntime` just starts/stops the coordinator in dependency order alongside the instance registry and processing scheduler — the same way it treats the others — instead of hand-managing a `ScheduledFuture`.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntime.kt
```diff
@@ -0,0 +1,170 @@
...
+        if (started) return
+
+        try {
+            try {
+                instanceRegistry.start()
+            } finally {
+                instanceStarted = instanceRegistry.isRunning
+            }
+
+            partitionCoordinator.rebalance()
```
Alek96: Redundant synchronous `rebalance()` at startup

`scheduleWithFixedDelay` starts the given `Runnable` as soon as possible, then applies the delay between the completion of one execution and the start of the next. So the scheduler already performs the initial rebalance on its own — the explicit synchronous `partitionCoordinator.rebalance()` immediately before it means the initial rebalance effectively runs twice around startup.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntime.kt
```diff
@@ -0,0 +1,170 @@
...
+    val outbox: Outbox,
+    private val instanceRegistry: OutboxInstanceRegistry,
+    private val partitionCoordinator: PartitionCoordinator,
+    private val processingScheduler: OutboxProcessingScheduler,
+    private val taskScheduler: TaskScheduler,
+    private val rebalanceInterval: Duration,
+    private val observationRegistry: () -> ObservationRegistry,
+    ownedPersistenceResources: List<AutoCloseable>,
+    ownedThreadingResources: List<AutoCloseable>,
+) : AutoCloseable {
```
Alek96, line 32: `OutboxRuntime` should separate "pause/resume scheduled jobs" from "dispose resources"

`OutboxRuntime` currently conflates two different lifecycle concerns behind a one-shot `start()` / `close()`:

- `start()` is guarded by a `started`/`closed` flag and returns early if already started;
- `close()` is terminal — it stops the components **and** closes owned persistence/threading resources, and the runtime can't be restarted afterward.

The problem is that the three underlying components — `OutboxInstanceRegistry` (phase 0), `OutboxProcessingScheduler` (phase 1), and (after the other comment) `PartitionCoordinator` — are all `SmartLifecycle`, and each scheduled job is meant to be individually **pausable and resumable** (e.g. during testing: pause processing, assert, resume). Spring's `SmartLifecycle` even exposes `default boolean isPauseable() { return true; }`, so `stop()`/`start()` are expected to be repeatable pause/resume operations, not a one-way teardown.

Because `OutboxRuntime` treats a `stop()` as part of a terminal `close()`, there's no clean way to pause the runtime's scheduled jobs and resume them without disposing the whole graph. "Pause" (stop scheduled jobs, keep the graph and resources alive) and "dispose" (tear everything down) are two distinct lifecycles that are currently mixed.

I'd suggest `OutboxRuntime` implement **both** interfaces, letting Spring's bean lifecycle handling separate the two concerns (see the [Spring bean nature docs on `Lifecycle`/`SmartLifecycle` vs. destruction callbacks](https://docs.spring.io/spring-framework/reference/core/beans/factory-nature.html)):

- **`SmartLifecycle`** — `start()` / `stop()` delegate (in phase/dependency order) to each component's own `SmartLifecycle`, giving repeatable pause/resume of the scheduled jobs. `isRunning()` reflects the components' running state rather than a one-shot flag.
- **`AutoCloseable`** — `close()` only disposes the *owned* resources (the persistence/threading `ownedResources`) at the very end, after the runtime is stopped. This keeps resource teardown as the terminal, one-time step, distinct from pausing.

To keep `close()` safe to call directly (without requiring a prior `stop()`), `close()` should first call `stop()` if the runtime hasn't already been stopped, and then dispose the owned resources. That preserves the "dispose is terminal" semantics while making the pause and dispose paths compose cleanly.

A couple of things to nail down if we go this way:

- **Start/stop ordering vs. phases:** resume should start in the same dependency order start uses today (registry → rebalance → processing), and stop should reverse it. Worth deciding whether we lean on `getPhase()` for this or keep the explicit ordering `OutboxRuntime` has now.
- **Idempotency & exception handling:** `stop()` and `close()` should both be idempotent (repeated calls are no-ops).

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/config/OutboxCoreInfrastructureAutoConfiguration.kt
rolandbeisel: Addressed in `7b1dfff`: added `@ConditionalOnSingleRuntimeMode` and applied it consistently across the Core, persistence, actuator, metrics, and observability auto-configurations.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/handler/assembly/HandlerRegistrationAssembler.kt
rolandbeisel: Addressed in `7b1dfff`: renamed the routing seam to `handlerSelector` across `OutboxHandlerInfrastructure`, `OutboxHandlerRegistrar`, and `HandlerRegistrationAssembler`. The KDoc now clarifies that selecting the primary method includes or excludes the complete handler registration.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntime.kt
rolandbeisel: Addressed in `7b1dfff`: the synchronous initial rebalance remains so that it is guaranteed to complete before processing starts. The periodic task now uses the three-argument `scheduleWithFixedDelay` overload and schedules its first run one rebalance interval later, avoiding the duplicate startup rebalance.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntimeResources.kt
```diff
@@ -0,0 +1,26 @@
...
+ * @property heartbeatScheduler Scheduler for instance heartbeats
+ * @property ownedResources Threading resources owned exclusively by the runtime
+ *
+ * @author Roland Beisel
+ * @since 1.10.0
+ */
+data class OutboxRuntimeResources(
+    val taskExecutor: TaskExecutor,
+    val taskScheduler: TaskScheduler,
+    val heartbeatScheduler: TaskScheduler,
```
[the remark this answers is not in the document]
rolandbeisel: I do not think object identity is a sufficient validity check here: sharing can be safe when the supplied scheduler has enough execution capacity, while a shared single-thread scheduler would indeed starve heartbeats during processing. The actual invariant is independent heartbeat execution capacity. I am leaving this thread open while we decide whether the runtime should guarantee that isolation internally or state it as an explicit contract for borrowed resources.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntimeFactory.kt
```diff
@@ -0,0 +1,115 @@
...
+                partitionAssignmentRepository = persistence.partitionAssignmentRepository,
+                partitionAssignmentCache = PartitionAssignmentCache(persistence.partitionAssignmentRepository),
+                clock = spec.clock,
+            )
+        val processorChain = createProcessorChain(handlers, spec)
+        val trigger = OutboxPollingTriggerFactory.create(properties, spec.clock)
+        val processingScheduler =
+            OutboxProcessingScheduler(
+                trigger = trigger,
+                taskScheduler = resources.taskScheduler,
```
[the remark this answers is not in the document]
rolandbeisel: Addressed in 35861a8: PartitionCoordinator now serializes rebalancing and complete processing batches through the same assignment lock, so custom multi-threaded TaskScheduler implementations cannot overlap them.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntime.kt
```diff
@@ -0,0 +1,170 @@
...
+        scheduledRebalance?.let { task -> attempt { task.cancel(false) } }
+        scheduledRebalance = null
+
+        if (instanceStarted) {
+            attempt(instanceRegistry::stop)
+            instanceStarted = false
+        }
+
+        persistenceResources.asReversed().forEach { resource -> attempt(resource::close) }
+        threadingResources.asReversed().forEach { resource -> attempt(resource::close) }
```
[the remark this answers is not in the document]
rolandbeisel: Addressed in 35861a8: OutboxRuntime now stops processing, partition coordination, and instance registration before closing owned threading resources, and closes persistence resources last. PartitionCoordinator.stop also waits for an active rebalance and suppresses a queued callback after stop.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntime.kt
rolandbeisel: Addressed in 35861a8: PartitionCoordinator now owns initial and periodic rebalance scheduling as SmartLifecycle phase 1. OutboxProcessingScheduler moved to phase 2, while OutboxRuntime only coordinates component start and stop.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntime.kt
rolandbeisel: We are keeping OutboxRuntime as the terminal AutoCloseable boundary. Its owned resources make close intentionally one-shot; channel pause and resume is not a current requirement, and a removed channel can be closed and recreated. The reusable components still use SmartLifecycle for dependency ordering.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntimeResources.kt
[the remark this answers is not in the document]
rolandbeisel: Addressed the contract in 35861a8: heartbeat execution must have capacity independent of long processing batches. The default remains a separate heartbeat scheduler; a shared scheduler is allowed only when its pool provides that capacity. Scheduler object identity is therefore not rejected.

namastack-outbox-core/src/main/kotlin/io/namastack/outbox/runtime/OutboxRuntime.kt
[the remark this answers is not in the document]
rolandbeisel: Completed in 50b5f1b: OutboxInstanceRegistry now serializes heartbeat execution and stop as well. Stop waits for an active heartbeat, and a queued callback observes the stopped state before touching persistence, including when the scheduler is borrowed rather than owned.

rolandbeisel:
@Alek96 I moved this PR back to draft because the latest Codex review found two valid P1 concurrency issues. I will be on vacation for the rest of this week and all of next week. If you have time and want to continue, please feel free to implement and push the fixes directly to `feature/gh-449-outbox-runtime`.

Suggested scope and acceptance criteria:

1. [Preserve the shutdown timeout](https://github.com/namastack/namastack-outbox/pull/453#discussion_r3958144012): keep processing and rebalance mutually exclusive, but `PartitionCoordinator.stop()` must not wait on a processing batch after `OutboxProcessingScheduler` has exhausted its shutdown timeout. It should stop future rebalances promptly and wait only for rebalance work that has actually begun. A queued rebalance blocked behind processing must observe the stopped state and exit without running. An equivalent design that preserves both guarantees is fine.
2. [Handle partial executor submission](https://github.com/namastack/namastack-outbox/pull/453#discussion_r3958144025): if the `TaskExecutor` accepts some record keys and then rejects a later submission, retain the original failure but wait for every accepted task before `processBatch` returns or rethrows. The assignment guard and processing lifecycle must remain active until those tasks finish.


Use `gh stack sync` for stack #456 (`#450 -> #453 -> #454`).

PR #454, opened 2026-09-02
3. JDBC Persistence Factory

rolandbeisel:


namastack-outbox-jdbc/src/main/kotlin/io/namastack/outbox/DefaultJdbcTableNameResolver.kt
```diff
@@ -14,22 +14,20 @@ import io.namastack.outbox.config.JdbcOutboxConfigurationProperties
  * @since 1.0.0
  */
 class DefaultJdbcTableNameResolver(
-    private val properties: JdbcOutboxConfigurationProperties,
+    properties: JdbcOutboxConfigurationProperties,
 ) : JdbcTableNameResolver {
```
Alek96, line 18: **`DefaultJdbcTableNameResolver` and `JdbcOutboxTableNamespace` are largely redundant**

After the namespace refactor, `DefaultJdbcTableNameResolver` no longer has any behavior of its own — it just constructs a `JdbcOutboxTableNamespace` from `JdbcOutboxConfigurationProperties` and delegates all three fields to it. It isn't a proxy or a wrapper; it's effectively a named constructor call. It's also referenced in exactly one place: `JdbcOutboxAutoConfiguration.jdbcTableNameResolver(properties)` (behind `@ConditionalOnMissingBean`). So we now have two classes implementing `JdbcTableNameResolver` that produce identical output for the same inputs.

**For simplifying:**
- Removes a redundant type — a reader currently has to discover that the two resolvers are equivalent.
- Single source of truth: `JdbcOutboxTableNamespace` already *is* the resolver; the default path should just use it.
- The properties → namespace mapping is a wiring concern that fits naturally in the auto-config.

**Against / to be careful about:**
- `DefaultJdbcTableNameResolver` is public and `@since 1.0.0`, and the `JdbcTableNameResolver` interface KDoc names it as *the* default implementation. Removing or changing it is a breaking change for anyone referencing it directly, so we'd want a deprecation path or at least a doc update pointing at the replacement.
- Coupling direction matters: `JdbcOutboxTableNamespace` is deliberately config-agnostic (plain params), which is what makes it reusable by the programmatic factory. Whatever we pick, the `properties` coupling should stay in the auto-config layer and not leak into the reusable namespace type.

**Possible solutions (pick whichever fits the API-stability stance):**

1. **Delete `DefaultJdbcTableNameResolver`; construct the namespace in `JdbcOutboxAutoConfiguration`.**
   The default bean method builds a `JdbcOutboxTableNamespace` directly from `properties`. `JdbcOutboxTableNamespace` becomes the single concrete resolver for both the auto-config and programmatic paths. Cleanest end state, but the biggest breaking change (the public `@since 1.0.0` type disappears).

   ```kotlin
   // DefaultJdbcTableNameResolver deleted

   // JdbcOutboxAutoConfiguration
   @Bean
   @ConditionalOnMissingBean
   internal fun jdbcTableNameResolver(
       properties: JdbcOutboxConfigurationProperties,
   ): JdbcTableNameResolver =
       JdbcOutboxTableNamespace(
           schemaName = properties.schemaName,
           tablePrefix = properties.tablePrefix,
           recordTableName = properties.tableNames.record,
           instanceTableName = properties.tableNames.instance,
           partitionTableName = properties.tableNames.partition,
       )
   ```

2. **Move `JdbcOutboxTableNamespace`'s internals into `DefaultJdbcTableNameResolver`, delete `JdbcOutboxTableNamespace`, and construct `DefaultJdbcTableNameResolver` in the auto-config like (1).**
   Keeps the historically public type name (`DefaultJdbcTableNameResolver`) and drops the newer one instead. The trade-off is that the resolver would then need the raw namespace params (schema/prefix/base names) rather than `properties`, so the programmatic factory would use this class directly — which is fine as long as it stays config-agnostic.

   ```kotlin
   // JdbcOutboxTableNamespace deleted; its logic now lives here
   class DefaultJdbcTableNameResolver(
       schemaName: String? = null,
       tablePrefix: String = "",
       recordTableName: String = "outbox_record",
       instanceTableName: String = "outbox_instance",
       partitionTableName: String = "outbox_partition",
   ) : JdbcTableNameResolver {
       override val outboxRecord: String = resolve(schemaName, tablePrefix, recordTableName)
       override val outboxInstance: String = resolve(schemaName, tablePrefix, instanceTableName)
       override val outboxPartitionAssignment: String = resolve(schemaName, tablePrefix, partitionTableName)

       private fun resolve(schema: String?, prefix: String, base: String): String {
           val table = "$prefix$base"
           return schema?.let { "$it.$table" } ?: table
       }
   }

   // JdbcOutboxAutoConfiguration
   @Bean
   @ConditionalOnMissingBean
   internal fun jdbcTableNameResolver(
       properties: JdbcOutboxConfigurationProperties,
   ): JdbcTableNameResolver =
       DefaultJdbcTableNameResolver(
           schemaName = properties.schemaName,
           tablePrefix = properties.tablePrefix,
           recordTableName = properties.tableNames.record,
           instanceTableName = properties.tableNames.instance,
           partitionTableName = properties.tableNames.partition,
       )
   ```

3. **Like (2), but add a dedicated factory/secondary constructor on `DefaultJdbcTableNameResolver` that maps `properties` → the class.**
   `DefaultJdbcTableNameResolver` exposes both a plain namespace-params constructor (for programmatic use) and a `properties`-based factory constructor (used by the auto-config). This keeps a single public resolver type, preserves the name, and cleanly separates the config mapping into the factory constructor — at the cost of the resolver knowing about `JdbcOutboxConfigurationProperties` again.

   ```kotlin
   class DefaultJdbcTableNameResolver(
       schemaName: String? = null,
       tablePrefix: String = "",
       recordTableName: String = "outbox_record",
       instanceTableName: String = "outbox_instance",
       partitionTableName: String = "outbox_partition",
   ) : JdbcTableNameResolver {
       override val outboxRecord: String = resolve(schemaName, tablePrefix, recordTableName)
       override val outboxInstance: String = resolve(schemaName, tablePrefix, instanceTableName)
       override val outboxPartitionAssignment: String = resolve(schemaName, tablePrefix, partitionTableName)

       private fun resolve(schema: String?, prefix: String, base: String): String {
           val table = "$prefix$base"
           return schema?.let { "$it.$table" } ?: table
       }

       companion object {
           fun fromProperties(properties: JdbcOutboxConfigurationProperties) =
               DefaultJdbcTableNameResolver(
                   schemaName = properties.schemaName,
                   tablePrefix = properties.tablePrefix,
                   recordTableName = properties.tableNames.record,
                   instanceTableName = properties.tableNames.instance,
                   partitionTableName = properties.tableNames.partition,
               )
       }
   }

   // JdbcOutboxAutoConfiguration
   @Bean
   @ConditionalOnMissingBean
   internal fun jdbcTableNameResolver(
       properties: JdbcOutboxConfigurationProperties,
   ): JdbcTableNameResolver =
       DefaultJdbcTableNameResolver.fromProperties(properties)
   ```

